PR #5, opened 2022-05-27
Another principle

lknik:
Loosely based on principles, too.

### DIFF
--- a/principles/index.html
+++ b/principles/index.html
@@ -81,6 +81,20 @@
         </p>
       </div>
     </section>
+    <section>
+      <h2>Leaks should be accounted for</h2>
+      <p>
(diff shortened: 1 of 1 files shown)


michaelkleber:
This was marked "call-for-consensus", but its proposed text includes the paragraph "Expand a little", so it is not ripe for merging in its current form.

AramZS:
@lknik Can you revise?

lknik:
@AramZS Thanks for pinging! Now I wonder if it really needs expanding, considered the overall document as of now. Would you indeed consider explaining the need for this rationale (as that was the "expand a little" part)? Otherwise, perhaps it's fine as is?

PR #6, opened 2022-05-28
Adding myself

lknik:


principles/index.html
```diff
@@ -15,6 +15,13 @@
           companyURL: 'https://nytimes.com/',
           url: 'https://berjon.com/',
           w3cid: 34327,
+        },
+                {
+          name: 'Lukasz Olejnik',
+          company: 'Independent researcher (Invited Expert)',
+          companyURL: 'https://lukaszolejnik.com.com/',
```
seanturner, line 22: I think there's an extra ".com"?
```suggestion
          companyURL: 'https://lukaszolejnik.com/',
```

principles/index.html
```diff
@@ -15,6 +15,13 @@
           companyURL: 'https://nytimes.com/',
           url: 'https://berjon.com/',
           w3cid: 34327,
+        },
+                {
+          name: 'Lukasz Olejnik',
+          company: 'Independent researcher (Invited Expert)',
```
seanturner, line 21: I'm going to suggest that we drop "Invited Expert" here because the CGs permit individual participation, i.e., there's really not an "Independent researcher" here.

```suggestion
          company: 'Independent researcher',
```

lknik:
Thanks for spotting. Indeed, right!

principles/index.html
lknik: Independent researcher, then, without IE. Thanks.

PR #12, opened 2022-09-22
Update index.html

chaals:
flesh out the point a bit

### DIFF
--- a/principles/index.html
+++ b/principles/index.html
@@ -66,7 +66,7 @@
   <section>
     <h2>Advertising-Specific Principles</h2>
     <section>
-      <h2>Consent Is Inappropriate</h2>
+      <h2>Uninformed Consent is Inappropriate</h2>
       <p>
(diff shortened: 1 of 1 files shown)


npdoty:
I've tried to incorporate the key point from this earlier text about consent into something more specific regarding measurement, and when/why consent would be inappropriate in that context, in #52.

While we may want separate text around consent generally, I suggest that we close this proposed change, at least for now. [Describing the conditions for consent being clear and informed is in the broader Privacy Principles document](https://w3ctag.github.io/privacy-principles/#consent-principles), as it's not at all advertising-specific.

npdoty:
As I noted last month, I think we should close this PR without merging, and instead continue with #52. If we need additional language, there could be a longer statement on advertising-specific consent for this doc, or if not advertising-specific, there might be suggestions to be made for the Privacy Principles doc.

martinthomson:
I would rather not perpetuate the idea that consent works and that the only reason it fails is because somehow people just didn’t understand.

PR #14, opened 2022-10-16
Private Computation Abstraction (MPC and TEEs)

eriktaubeneck:
This pull request attempts to add "private computation" as an abstraction, which can be (potentially) be achieved by both MPC and TEEs. This avoids the need to include a specific technology in the standard, and enables web platform vendors (i.e. implementers) to decide which constructions to support.

It also proposes that web platform vendors can choose their specific privacy budget allotment (which is somewhat tied to the ability to support multiple private computation construction.


threat-model/readme.md
```diff
@@ -230,11 +204,53 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
-1. _Unit of privacy_: The set of parties used to manage the privacy budget.
+4. _Aggregation minimum threshold_: the number of clients/users that need to be included in a given aggregation.
+
+We can divide these parameters into two groups: parameters which should be part of the standard (and thus require consensus) and parameters which can be configured by the web platform vendor. The first two parameters, _unit of privacy_, and _epoch_ should likely be part of the standard. However, the other two parameters, ε and _aggregation minimum threshold_ could differ across web platform vendors without compromising interoperability.
+
+Let's first address, ε. Suppose that we have _Browser A_ and _Mobile OS B_ which, respectively, decide that the appropriate budget is "1 unit/epoch" and "5 unit/epoch". Luckily, in the worst case, privacy budgets are additive, and thus can be split into smaller pieces. A user of this API could then decide to use "1 unit/epoch" on the API from both _Browser A_ and _Mobile OS B_, and then continue to use the remaining "4 unit/epoch" on the API from _Mobile OS B_.
```
martinthomson, line 216: I think that it might be worth expanding a little on how this division might be achieved in practice.  Something like:

> This might be achieved by using a system that distributes information from both _Browser A_ and _Mobile OS B_ to a private computation entity that is configured to consume 1 unit per epoch.  Information from _Mobile OS B_ is additionally distributed to another entity that is configured to consume the remaining 4 units.  As a practical matter, the two logical private computation entities here could be operated by the same organizations, but they would use different configurations.  For instance, each configuration might use different keying material.  In this case, _Browser A_ is able to limit information release to 1 unit per epoch by choosing where information can be sent.

threat-model/readme.md
```diff
@@ -230,11 +204,53 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+Secondly, let's address _aggregation minimum threshold_. Suppose that the same _Browser A_ and _Mobile OS B_ have also, respectively, decided that the _aggregation minimum threshold_ should be "X people" and "2X people". Because this is a minimum, when using the API from both, the minimum can be set to 2X, and satisfy both constraints. However, when using the API from only _Browser A_, the minimum could be reduced to X.
```
```suggestion
Secondly, let's address _aggregation minimum threshold_. Suppose that the same _Browser A_ and _Mobile OS B_ have also, respectively, decided that the _aggregation minimum threshold_ should be "X people" and "2X people". Because this is a minimum, when using the API from both, the minimum can be set to 2X, and satisfy both constraints. However, when using data that comes from _Browser A_ only, the minimum could be reduced to X.
```

threat-model/readme.md
```diff
@@ -230,11 +204,53 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+## 4. Private Computation Configurations
+
+Many of these proposals aim to leverage the idea of _private computation_, touched on briefly in the introduction. In it's ideal form, a private computation environment would allow for the evaluation of a predefined function (i.e., the constrained purpose,) without revealing any new information to any party beyond the output of that predefined function. This is commonly used to perform aggregation over inputs which, individually, must not be revealed. It is also often used to then apply differentially private noise to those aggregates.
+
+There are currently two proposed constructions of a private computation environment: _multi-party computation_ (MPC) and _trusted execution environments_ (TEEs). The following two subsections outline how these can be to create a private computation environment, how they fit into the threat model, and the assumptions required to achieve our goal of assuring that the inputs are not revealed.
```
martinthomson, line 224:
```suggestion
There are currently two proposed constructions of a private computation environment under consideration: _multi-party computation_ (MPC) and _trusted execution environments_ (TEEs). The following two subsections outline how these can be to create a private computation environment, how they fit into the threat model, and the assumptions required to achieve our goal of assuring that the inputs are not revealed.
```

threat-model/readme.md
```diff
@@ -230,11 +204,53 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+### 4.1 Multi-party Computation (MPC)
```
martinthomson, line 226:
```suggestion
### 4.1 Multi-party Computation
```

threat-model/readme.md
```diff
@@ -230,11 +204,53 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+Multi-party computation is a cryptographic protocol in which distinct parties can collectively operate on data which remains oblivious to any individual party throughout the computation, but allows for joint evaluation of a predefined function.
+
+These protocols typically work with data which is _secret shared_. For example, a three way secret share of a value v = s<sub>1</sub> + s<sub>2</sub> + s<sub>3</sub> can be constructed by generating two random values for s<sub>1</sub> and s<sub>2</sub>, and then computing s<sub>3</sub>. At this point, each value s<sub>i</sub> individually appears random, and thus v remains oblivious even when a party learns all by one s<sub>i</sub>. Note that secret shares can also be generated using XOR with random byte strings.
```
martinthomson, line 230:
```suggestion
These protocols typically work with data which is _secret shared_. For example, a three way _additive_ secret share of a value v = s<sub>1</sub> + s<sub>2</sub> + s<sub>3</sub> can be constructed by generating two random values for s<sub>1</sub> and s<sub>2</sub>, and then computing s<sub>3</sub> = v - s<sub>1 - s<sub>2</sub>. At this point, each value s<sub>i</sub> individually appears random, and thus v remains oblivious as long as no single entity learns all values of s<sub>i</sub>. A similar secret sharing schemes uses XOR in place of addition; alternatively, Shamir's secret sharing uses polynomial interpolation.
```

threat-model/readme.md
```diff
@@ -230,11 +204,53 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+In terms of our threat model, these parties are aggregators and we assume that an attacker can control some subset of those aggregators. That exact threshold may be different for a given proposal, for example, we may assume that an attacker can only control one out of three aggregators. This would enable, in cryptographic terms, a maliciously secure, honest two out of three majority MPC.
```
martinthomson, line 232: I think that we want to talk about MPC in the HBC model briefly and note that this is not suitable for this context as we assume that a subset of the entities that perform the computation are controlled by an attacker.

threat-model/readme.md
```diff
@@ -230,11 +204,53 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+Given these aggregators, a _client/user_ is able to generate a secret sharing of their input data, and then securely communicate one secret share to each aggregator. This allows the aggregators to then perform the MPC protocol to compute the predefined function. The _client/user_ is assuming that an attacker cannot control enough aggregators to violate that protocol, which implies that they only perform the expected function.
```
martinthomson, line 234:
```suggestion
Given these aggregators, a _client/user_ is able to generate a secret sharing of their input data, and then securely communicate one secret share to each aggregator. This allows the aggregators to then perform the MPC protocol to compute the predefined function. The _client/user_ trusts that an attacker cannot control enough aggregators to violate that protocol.

If the protocol is faithfully executed, the aggregate result can then be reported to the intended recipient without the entities that performed the computation needing to witness the answer.

This threat model does not require that the MPC protocol provide safeguards against an aggregator spoiling the answers from the system. Though a spoiled answer is possible in many protocols, the primary threat that a compromised aggregator presents is to the privacy of _clients/users_. We assume that there is some contractual relationship between those requesting aggregation and the entities that perform that aggregation such that the aggregators lack significant incentive to spoil results.
```

threat-model/readme.md
```diff
@@ -230,11 +204,53 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+### 4.2 Trusted Execution Environments (TEEs)
+
+Trusted execution environments are specialized hardware where encrypted data can be sent "in" to an enclave where it is decrypted and operated on, but cannot be otherwise accessed. It also produces an "attestation" that provides a means of verifying that the code run on that data is the code that was intended. Different hardware manufacturers offer different types of TEEs, and there are various assumptions which need to be made about the manufacturer, hardware operator, and other tenants on the hardware in order to achieve our ideal of input data remaining oblivious. There are documented attacks for many of these types of hardware, and while we don't expect all web platform vendors to support this construction, some vendors have expressed comfort with the required assumptions.
```
martinthomson, line 238:
```suggestion
Trusted execution environments are specialized hardware where encrypted data can be sent "in" to an enclave where it is decrypted and operated on, but cannot be otherwise accessed. A TEE can produce an "attestation" that acts as a claim - backed by the vendor of the TEE - that only specific code is actively running.

Different hardware manufacturers offer different types of TEEs, and there are various assumptions which need to be made about the manufacturer, hardware operator, and other tenants on the hardware in order to achieve our ideal of input data remaining oblivious. There are documented attacks for many of these types of hardware, and while we don't expect all web platform vendors to support this construction, some vendors have expressed comfort with the required assumptions.
```

threat-model/readme.md
```diff
@@ -230,11 +204,53 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+In order to utilize a TEE to achieve private computation, we need two properties. First, the data must only be able to be decrypted within the TEE, and second, we need to assure that the TEE only executes code which evaluates the predefined function.
+
+In a very simplified form, we could imagine there is one TEE which is used by every _first party_/_delegated party_ to run the predefined function. In this case, the _user/client_ could encrypt their data with the public key of the TEE. However, this ignores assuring the code running is that that's intended. The TEE produces an attestation of the code it's running, however the _client/user_ needs that attestation to be verified before their input data is decrypted.
```
martinthomson, line 242: I don't think that this adds much here.

You could imagine a singular TEE like this working as follows:

1. The TEE mints a keypair and publishes an attestation that covers the public key (and the software that it is running. Clients are configured with this public key after the attestation is validated.
2. Clients encrypt data toward this public key.
3. Record collectors instantiate (a copy of?) the TEE and submit their encrypted data to it for computation.

This would work, but it wouldn't scale.  It also creates some really awkward scaling questions.  I don't know what the general attitude is toward cloning a TEE, but if that were possible, then that would be bad for privacy as the same queries could be repeated.

I think it's better just to describe the envisaged deployment model directly, as above.

threat-model/readme.md
```diff
@@ -230,11 +204,53 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+This can be achieved by utilizing an aggregator who verifies the TEE attestation on behalf of the _client/user_. In this case, the _client/user_ encrypts the data towards the aggregator, and the aggregator provides the decryption key into the TEE if and only if the attestation verifies the TEE is only running the expected code. This still has one problem, however, as we assume that our attacker can compromise at least one aggregator.
```
martinthomson, line 244: Michael is using "coordinator" here rather than "aggregator".  The TEE is an aggregator here as well.

threat-model/readme.md
```diff
@@ -230,11 +204,53 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+We can extend this one step further, by utilizing not one, but multiple aggregators. Each aggregator can hold part of a _threshold key_, which allows data to be encrypted such that it requires all (or a predefined subset) of keys to decrypt. Each aggregator verifies the attestation and provides their decryption key into the TEE if and only if the attestation verifies the TEE is only running the expected code. Thus, only the TEE is ever able to decrypt the data, so long as the attacker is only able to control a subset of aggregators.
+
+### 4.3 Abstracting Private Computation
+
+Both of these constructions rely on a set of aggregators, where we assume that an attacker is unable to control all aggregators. Let a specific set of aggregators be called an aggregator network. Each web platform vendor will need to make a judgment as to if it's reasonable to assume that an aggregator network can be _trusted_, e.g. that attackers would be unable to compromise all aggregators within that specific aggregator network.
```
martinthomson, line 250:
```suggestion
Both of these constructions rely on a set of aggregators, where we assume that an attacker is unable to control all aggregators. Let a specific set of aggregators be called an aggregator network. Each web platform vendor will need to make a judgment as to if it's reasonable to assume that an aggregator network can be _trusted_.  That is, that attackers would be unable to compromise any subset of the aggregators within that specific aggregator network such that the privacy guarantees of the system could not be met.
```

threat-model/readme.md
```diff
@@ -230,11 +204,53 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+To make this judgment, web platform vendors are likely to consider various properties about helper party networks such as diversity in ownership of the company/organization operating the aggregators, diversity in cloud provider (if relevant) used by the aggregators, diversity in jurisdictions in which the aggregators operate, etc.
+
+It's unlikely that all web vendors will arrive at the same decision as to which aggregator networks can be assumed to be trusted. Our aim, however, is for many such aggregator networks to exist that are trusted by most web platforms, allowing for choice and competition for the first/delegated parties that leverage these APIs. Moreover, as we may have web platform vendors providing different privacy budgets (see section [3.3 Privacy Parameters](#33-privacy-parameters),) the first/delegated parties may use multiple aggregator networks to utilize those different budgets.
```
martinthomson, line 254: I'm not sure that I'm willing to give up on us reaching a better agreement here.  I'd prefer if we hedge this a little more carefully.

I would like to see us agree on a core set of methods and operators so that sites can get a uniform service across all platforms.  Some platforms might offer increments over that, with additional privacy budget allowances or different modes of operation, but if we can't agree on the common subset, then I won't personally be happy with the outcome.  That very much creates an incentive for sites to prefer certain browsers over others, a situation which is already pretty acute for those browsers that aren't Chrome.

martinthomson (changes requested):
Lots of comments here.  I'm not happy with the tail end of this, but I'm running out of steam for doing suggestions.

The build-up is fairly solid though.  I've some points on framing throughout.

I couldn't identify where to put the DP piece in; that might need some more thought.

threat-model/readme.md
```diff
@@ -10,16 +10,18 @@ This document is currently a **draft**, submitted to the PATCG (and eventually t
...
+Many of these proposals attempt to leverage the concept of _private computation_ as a component of these purpose-constrained APIs. An ideal private computation environment would allow for the evaluation of a predefined function (i.e., the constrained purpose,) without revealing any new information to any party beyond the output of that predefined function to the specified parties. This is commonly used to perform aggregation over inputs which, individually, must not be revealed. It is also often used to then apply differentially private noise to those aggregates.
 
-In the presence of this adversary, APIs should aim to achieve the following goals:
+Private computation has various constructions, each with different assumptions. The two primary forms considered by existing proposals are _multi-party computation_ (MPC) and _trusted execution environments_ (TEEs.) MPC relies on distinct parties, or _aggregators_, who perform a cryptographic protocol, while TEEs rely on specialized hardware that provides isolation for computation on sensitive data.
```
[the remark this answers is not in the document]
eriktaubeneck: I thought about including FHE, though I left it out for 2 reasons:
1) it's not clear to me how it could actually work in this setting (typically the party that encrypts is the one that also decrypts, which is difficult in a setting with distributed clients)
2) I wanted to focus on the specific constructions (instantiations?) that have been proposed thus far.

It's also not obvious (at this point) that TEEs require two entities.

threat-model/readme.md
```diff
@@ -159,6 +135,8 @@ An aggregator is an individual party which participates in an aggregation protoc
 
 If enough aggregators (beyond the proposal specific subset) collude e.g. by sharing unencrypted input shares), then none of the properties of the system hold. Such scenarios are outside the threat mode.
 
+However, we do assume that an attacker can always control at least one aggregator (i.e. there are no perfectly trusted aggregators.)
```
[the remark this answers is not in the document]
eriktaubeneck: This is intended to motivate the need for multiple coordinators with threshold keys (as otherwise, the coordinator is a single point party with the complete decryption key.) I agree this maybe not the right place to put this, and happy to think about this more, but I do think it's important that there isn't a single party which has the ability to fully decrypt the data.

For the cloud service provider, the claim here is that the TEE provides sufficient protection against intrusion that they are unable to access the data. (Different vendors have different opinions on if this is sufficient.)

For the software vendor, I'm not sure how this is distinct from the software used for an MPC implementation.

threat-model/readme.md
eriktaubeneck, line 216:
```suggestion
Let's first address, ε. Suppose that we have _Browser A_ and _Mobile OS B_ which, respectively, decide that the appropriate budget is "1 unit/epoch" and "5 unit/epoch". Luckily, in the worst case, privacy budgets are additive, and thus can be split into smaller pieces. A user of this API could then decide to use "1 unit/epoch" on the API from both _Browser A_ and _Mobile OS B_, and then continue to use the remaining "4 unit/epoch" on the API from _Mobile OS B_.

This might be achieved by using a system that distributes information from both Browser A and Mobile OS B to a private computation entity that is configured to consume 1 unit per epoch. Information from Mobile OS B is additionally distributed to another entity that is configured to consume the remaining 4 units. As a practical matter, the two logical private computation entities here could be operated by the same organizations, but they would use different configurations. For instance, each configuration might use different keying material. In this case, Browser A is able to limit information release to 1 unit per epoch by choosing where information can be sent.
```

threat-model/readme.md
eriktaubeneck:
```suggestion
In terms of our threat model, these parties are aggregators and we assume that an attacker can control some subset of those aggregators. That exact threshold may be different for a given proposal, for example, we may assume that an attacker can only control one out of three aggregators. This would enable, in cryptographic terms, a maliciously secure, honest two out of three majority MPC.

Another security model for MPC is _honest by curious_, where the input data remains oblivious so long as aggregators do not deviate from the protocol. Since we are assuming that an attacker can control some subset of the aggregators, it would be able to deviate from the protocol, and thus the _honest but curious_ is not suitable.
```

threat-model/readme.md
eriktaubeneck: Yes, I was hoping that we might be able to use the same term, since the threats are similar, and it allows for the nice abstraction of "browser just support different aggregator networks." 

Another approach would be to separately include "coordinators" as types of actors above, and then have a meta class of actors, something like "helper parties", of which both "coordinators" and "aggregators" are. 

> The TEE is an aggregator here as well.

The TEE itself isn't an actor, right? We do have the cloud provider (which upon reflection, we should add in general, as it has implications for both) and the TEE manufacturer (which may also need to be added.)

threat-model/readme.md
eriktaubeneck: I like that framing. Let me take another pass at this section.

eriktaubeneck:
Thanks Martin, very helpful feedback. Merged a bunch of the suggestions in, and will spend some more time continuing to iterate.

bmayd:
Is there a value in noting the potential for combining private computation methods i.e., TEEs and MPCs being leveraged for different aspects of workflows?

threat-model/readme.md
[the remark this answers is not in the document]
martinthomson: One design for FHE is to have the record collector (or someone they contract) perform the aggregation using ciphertext.  Then a trusted entity decrypts the output.

As for TEEs being multi-party, I found the comments by Nigel [here](https://github.com/patcg/meetings/issues/39#issuecomment-1084031361) (relayed by you as it happens) very helpful in articulating the high-level constraints.  Nigel's two parties are the client and the TEE operator, but as we've observed, this introduces a problem of scale that pushes toward the coordinator-based design described here.

threat-model/readme.md
martinthomson: I think that you are right in attempting to identify the entities that have a critical role in protecting the data.  For TEEs that is both the vendor and the operator.  I guess for an MPC, each node has to contend with operational and procedural issues, but we are generally comfortable lumping those together under a single umbrella as the same entity is on the hook for both, even if they subcontract aspects of the operation out.

martinthomson:
@bmayd, if we choose something other than TEE, then that technology could be used to fortify the operational security of any alternative.  I don't believe that this would be a strict requirement from our perspective as a prospective client of these services, but it is an option that becomes available if we are less than perfectly certain about the operational practices of a potential MPC participant (for example).

threat-model/readme.md
palenica: nit: s/honest by curious/honest but curious

threat-model/readme.md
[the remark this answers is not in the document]
palenica: For what it's worth, the recent release of privacy sandbox aggregation service uses two coordinators
https://github.com/privacysandbox/aggregation-service

palenica:
Nit: this doc talks briefly about privacy budgeting in the privacy section. It also talks about trust and security in the MPC and TEE private computation options, but only in the context of keeping the data and computation private. Would it make sense to explicitly cover how to make privacy budgeting trusted? 
Keeping un-replayable privacy budget state tends to require a trusted party (or multiple trusted parties who check on each other) as well as some sort of access control to prevent parties from consuming (or observing) each other's budgets.

bmayd:
From the introduction:

> In the presence of this adversary, APIs should aim to achieve the following goals:
-Privacy: Clients (and, more specifically, the vendors who distribute the clients) trust that (within the threat models), the API is purpose constrained. That is, all parties learn nothing beyond the intended result (e.g., a differentially private aggregation function computed over the client inputs.)
-Correctness: Parties receiving the intended result trust that the protocol is executed correctly. Moreover, the amount that a result can be skewed by malicious input is bounded and known.

I suggest instead of Privacy, we make the first bullet:

Purpose Limitation: User-agents are reasonably assured that the API is purpose constrained such that no party can acquire data outputs other than what is intended and expected by the user-agent, given the inputs it provides.

Add bullets for verifiable input and auditability:

Verifiable Input: Parties using the API are reasonably assured that data provided by user-agents is accurate, reliable and honest.

Auditability: Parties providing data to, or receiving data from, the API can receive reports identifying: when, how, by whom and to whom data was communicated; when, how and by whom data and processed.

Correctness: Parties receiving the intended result can verify that the protocol is executed correctly and that the amount a result can be skewed, intentionally by adding noise, or by malicious input, is bounded, known and reported.

threat-model/readme.md
eriktaubeneck: thanks @palenica!

threat-model/readme.md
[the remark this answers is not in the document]
eriktaubeneck: Ok, that works for me.

I've been turning over the multi-party idea (wrt this and other parts), and I realized the constraint I've been circling around is the idea that we likely want 2 (or more) parties who are *known up front* and can be trusted to reach some conditions (typically non-collusion.) This was my concern with the single coordinator model, as one of the parties is the report collector, who is not known up front. I additionally wasn't sure if the client/user would be considered one of these parties, but maybe that point is moot due to the scale.

I'm fine with this addition, and will work in the *known up front* part elsewhere.

threat-model/readme.md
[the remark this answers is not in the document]
eriktaubeneck: @palenica love to see it, thanks for pointing it out! I'm going to rework this whole section, and jump right to that design.

threat-model/readme.md
eriktaubeneck: > but we are generally comfortable lumping those together under a single umbrella as the same entity is on the hook for both, even if they subcontract aspects of the operation out.

Yes, but I think we aren't comfortable with two distinct entities running MPC within a single cloud provider.

martinthomson: Right.  Though some might be, not all of us are.  Still, it falls under the rubric of having some form of validation of the MPC operations as a whole.

threat-model/readme.md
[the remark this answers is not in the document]
eriktaubeneck: Weird GitHub bug - I think the comments appeared out of order because of comment time/review submission time inconsistencies... Anyway, my reply was in response to @martinthomson.

@marianapr agreed that the number of parties/non-collusion is a different type of assumption than the architechture and trust assumptions, and I can take another pass to try and call these two assumptions out distinctly. That said, this PR is attempting to suggest that those assumptions don't need to be made in the standard or threat model, but rather by the implementors (web platform vendors.)

threat-model/readme.md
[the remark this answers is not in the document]
eriktaubeneck: I thought about this, but the coordinators purpose is not only to keep the key secret, but also to validate the attestation. If that validation also happens in a TEE, I think the client/user would also want to be able to validate an attestation that *that* TEE is configured correctly, and it ends up turtles all the way down.

threat-model/readme.md
```diff
@@ -0,0 +1,360 @@
...
+2. Delegated parties can modify client assets through the proposed APIs, which may include overwriting/modifying data provided by the first party or other delegated parties.
+
+
+#### 1.3.3. Mitigations
+
+1. The same mitigations as Third Parties ([section 1.2.3](#123-Mitigations))
+2. The client should prevent the ability of a delegated party to overwrite/modify data provided by the first party or other delegated parties. If this is not possible, the client should allow the first party to control and limit a delegated party's ability in this manner. (See [section 2. First Parties, Embedded Parties and Delegated Parties](#2-First-Parties-Embedded-Parties-and-Delegated-Parties) for more details.)
+
+
+### 1.4 Helper Parties and Helper Party Networks
```
eriktaubeneck, line 112: I separated out distinct parties here for "Helper parties" (as a class), "aggregators" for helper parties in an MPC, and "coordinators" for helper parties supporting a TEE. I also added Cloud providers, TEE operators, and TEE manufacturers.

@martinthomson and @marianapr, if you'd be so kind to review again, it would be much appreciated!

threat-model/readme.md
```diff
@@ -0,0 +1,360 @@
...
+#### 1.8.3 Mitigations
+
+1. Helper parties networks should utilize sufficiently distinct cloud providers beyond the proposal specific subset which the attacker is assumed to control.
+
+
+### 1.9 Operators of TEEs
+
+As a piece of hardware, TEEs will have an operator with access to the machine. Most commonly, this will be a cloud provider. Depending on the specific hardware, there may be known vulnerabilities in which an attacker who only controls the operator can violate the obliviousness of client/user data. These attacks are outside this threat model, but are likely to inform specific web platform decisions about which instantiations of private computation to support.
+
+#### 1.9.1 Assets
```
eriktaubeneck, line 205: I wasn't quite sure how to approach this (and I'm unsure if the text immediately above is appropriate.) I'm not sure if we should write this out assuming the operator can actually see inside the TEE, and then call out the hardware as a mitigation? Any suggestions here would certainly be appreciated! (cc @martinthomson @marianapr @palenica)

threat-model/readme.md
```diff
@@ -0,0 +1,356 @@
...
+
+Arbitrary helper party networks will not be able to automatically participate in these protocols, as web platform vendors will need to provide some form of authorization. As such, our goal is to reach consensus on a core set of methods and helper party networks so that first/delegated parties can get uniform service across web platform vendors.
+
+Beyond this core set of methods and operators, we aim to make these APIs interoperable such that individual web platform vendors may offer increments over that, by authorizing a large set of helper party networks, certain instantiations of private computation, and extended privacy budgets. This would allow first/delegated parties to layer their usage of these APIs, utilizing that core set of methods and helper party networks for uniform service, and layering on incremental service where it's available.
```
eriktaubeneck, line 356: @martinthomson updated WRT your feedback here: https://github.com/patcg/docs-and-reports/pull/14#discussion_r997710575

threat-model/readme.md
```diff
@@ -0,0 +1,356 @@
...
+
+The currently proposed system utilizes a helper party network composed of coordinators. This helper party network participate in a key exchange protocol to generate a _threshold key pair_, with a public key for encryption and several partial private keys such that each coordinator has a single partial private key, and all (or a predefined subset) are required for decryption. This allows the _client/user_ to encrypt the data towards the helper party network.
+
+When a TEE is instantiated to perform an aggregation, each coordinator in the helper party network can validate an attestation from the TEE and provide their partial private key if and only if the attestation verifies that the TEE is only running the expected code. Thus, only the TEE is ever able to decrypt the data, so long as the attacker is only able to control a subset of coordinators in the helper party network.
+
+Note that in [section 1.7. Helper party collusion](#17-Helper-party-collusion), we assume that at least one helper party in a network can be controlled by the attacker, thus requiring at least two coordinators in the helper party network.
```
eriktaubeneck, line 346: @martinthomson updated with your feedback here: https://github.com/patcg/docs-and-reports/pull/14#discussion_r997709062

cc @marianapr and @palenica

threat-model/readme.md
martinthomson: I think that we want to be direct about the threat model that a TEE is developed under.  That is, physical access to the hardware might, with some non-trivial effort, be used to access the active state, which would otherwise be protected.

And we have to consider that the operator has physical access.  This is precisely why some of us find use of a TEE uncomfortable/unacceptable.

I don't think we win by prevaricating on this point.  We win by acknowledging it and tackling it head-on.  That means strict requirements on operational practices by TEE operators, audits, and other such non-technical measures.

threat-model/readme.md
```diff
@@ -99,85 +93,159 @@ See [section 2. First Parties, Embedded Parties and Delegated Parties](#2-First-
...
+TODO
+
+
+### 1.10 TEE Manufacturers
+
+TEEs can provide "attestation" which verifies that the TEE is running in the expected state and running the expected code. These attestations are typically produced with an asymmetric key, where the private key is physically embedded into the TEE, and the public key is published via a system similar to a certificate authority.
+
+#### 1.9.1 Assets
+
+1. Private keys embedded in TEEs and their corresponding public keys.
```
martinthomson, line 223: It is quite possible that operational practices at the TEE manufacturer is good enough that they don't have access to the private keys, but they will have access to the "root" key.

We should assume that they have both.

threat-model/readme.md
```diff
@@ -230,11 +294,63 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+To make this judgment, web platform vendors are likely to consider various properties about helper party networks such as diversity in ownership of the company/organization operating the helper parties, diversity in cloud provider (if relevant) used by the helper parties, diversity in jurisdictions in which the helper parties, cloud providers, and TEE operators operate, etc. The specific instantiation of private computation will also be a factor in this decision.
+
+Arbitrary helper party networks will not be able to automatically participate in these protocols, as web platform vendors will need to provide some form of authorization. As such, our goal is to reach consensus on a core set of methods and helper party networks so that first/delegated parties can get uniform service across web platform vendors.
```
martinthomson, line 354:
```suggestion
Arbitrary helper party networks will not be able to automatically participate in these protocols, as web platform vendors will need to provide some form of authorization. As such, our goal is to reach consensus on a core set of methods and helper party networks so that first/delegated parties can get uniform service across web platform vendors. This will in turn allow sites/apps to access a basic set of attribution capabilities across all participating platforms.
```

threat-model/readme.md
```diff
@@ -230,11 +294,63 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+Beyond this core set of methods and operators, we aim to make these APIs interoperable such that individual web platform vendors may offer increments over that, by authorizing a large set of helper party networks, certain instantiations of private computation, and extended privacy budgets. This would allow first/delegated parties to layer their usage of these APIs, utilizing that core set of methods and helper party networks for uniform service, and layering on incremental service where it's available.
```
martinthomson, line 356:
```suggestion
Beyond this core set of methods and operators, we aim to make these APIs in a way that allows individual web platform vendors to offer increments over that.  This might be by authorizing a large set of helper party networks, alternative instantiations of private computation, or by extending privacy budgets. This would allow first/delegated parties to layer their usage of these APIs, utilizing that core set of methods and helper party networks for uniform service, and layering on incremental service where it's available.
```

martinthomson (approved):
A solid start.  More comments, but I think that you should consider merging and filing some issues for any shortfall.  Not long until we next meet and you want to give people time to digest this.

threat-model/readme.md
```diff
@@ -99,85 +93,159 @@ See [section 2. First Parties, Embedded Parties and Delegated Parties](#2-First-
...
+### 1.5. Aggregator Helper Party
+
+An aggregator is type of helper party which participates in a helper party network to instantiate an MPC based private computation system. Aggregators receive secret shares of the input data and interact with the other aggregators in the helper party network to perform the aggregation. See [section 4.1 Multi-pary Computation](#41-Multi-party-Computation) for more details.
 
-1. Unencrypted individual share.
-2. All the assets from first and third parties.
-3. Shares of the output of the aggregation protocol.
-4. Identity of other aggregators.
 
+#### 1.5.1. Assets
```
csharrison, line 126: Maybe we can add this in a follow-up but something missing in bot the aggregator assets and coordinator assets is the _state_ of the privacy budget accountant which you reference later on in the doc.

threat-model/readme.md
```diff
@@ -230,11 +294,63 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+Let's first address, ε. Suppose that we have _Browser A_ and _Mobile OS B_ which, respectively, decide that the appropriate budget is "1 unit/epoch" and "5 unit/epoch". Luckily, in the worst case, privacy budgets are additive, and thus can be split into smaller pieces. A user of this API could then decide to use "1 unit/epoch" on the API from both _Browser A_ and _Mobile OS B_, and then continue to use the remaining "4 unit/epoch" on the API from _Mobile OS B_.
+
+This might be achieved by using a system that distributes information from both Browser A and Mobile OS B to a private computation entity that is configured to consume 1 unit per epoch. Information from Mobile OS B is additionally distributed to another entity that is configured to consume the remaining 4 units. As a practical matter, the two logical private computation entities here could be operated by the same organizations, but they would use different configurations. For instance, each configuration might use different keying material. In this case, Browser A is able to limit information release to 1 unit per epoch by choosing where information can be sent.
```
csharrison, line 308:
```suggestion
This might be achieved by using a system that distributes information from both Browser A and Mobile OS B to a private computation entity that is configured to consume 1 unit per epoch. Information from Mobile OS B is additionally distributed to another entity that is configured to consume the remaining 4 units. As a practical matter, the two logical private computation entities here could be operated by the same organizations, but they would use different configurations. These configurations could be chosen via browser choice of keying material, or directly specified in individual user input.
```
I am nervous about complicating PKI even more. Let's leave our options open for specifying the configuration.

threat-model/readme.md
```diff
@@ -230,11 +294,63 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+Secondly, let's address _aggregation minimum threshold_. Suppose that the same _Browser A_ and _Mobile OS B_ have also, respectively, decided that the _aggregation minimum threshold_ should be "X people" and "2X people". Because this is a minimum, when using the API from both, the minimum can be set to 2X, and satisfy both constraints. However, when using data that comes from _Browser A_ only, the minimum could be reduced to X.
+
+## 4. Private Computation Configurations
+
+Many of these proposals aim to leverage the idea of _private computation_, touched on briefly in the introduction. In it's ideal form, a private computation environment would allow for the evaluation of a predefined function (i.e., the constrained purpose,) without revealing any new information to any party beyond the output of that predefined function. This is commonly used to perform aggregation over inputs which, individually, must not be revealed. It is also often used to then apply differentially private noise to those aggregates.
```
csharrison, line 314:
```suggestion
Many of these proposals aim to leverage the idea of _private computation_, touched on briefly in the introduction. In it's ideal form, a private computation environment would allow for the evaluation of a predefined function (i.e., the constrained purpose,) without revealing any new information to any party beyond the output of that predefined function. This is commonly used to run a privacy mechanism across user input, which individually must not be revealed.
```
I think we can be more general here.

threat-model/readme.md
```diff
@@ -230,11 +294,63 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+### 4.1 Multi-party Computation
+
+Multi-party computation is a cryptographic protocol in which distinct parties can collectively operate on data which remains oblivious to any individual party throughout the computation, but allows for joint evaluation of a predefined function.
+
+These protocols typically work with data which is _secret shared_. For example, a three way _additive_ secret share of a value v = s<sub>1</sub> + s<sub>2</sub> + s<sub>3</sub> can be constructed by generating two random values for s<sub>1</sub> and s<sub>2</sub>, and then computing s<sub>3</sub> = v - s<sub>1 - s<sub>2</sub>. At this point, each value s<sub>i</sub> individually appears random, and thus v remains oblivious as long as no single entity learns all values of s<sub>i</sub>. A similar secret sharing schemes uses XOR in place of addition; alternatively, Shamir's secret sharing uses polynomial interpolation.
```
csharrison, line 322:
```suggestion
These protocols typically work with data which is _secret shared_. For example, a three way _additive_ secret share of a value v = s<sub>1</sub> + s<sub>2</sub> + s<sub>3</sub> can be constructed by generating two random values for s<sub>1</sub> and s<sub>2</sub>, and then computing s<sub>3</sub> = v - s<sub>1 - s<sub>2</sub>. At this point, each value s<sub>i</sub> individually appears random, and thus v remains oblivious as long as no single entity learns all values of s<sub>i</sub>. A similar secret sharing schemes uses XOR in place of addition; alternatively, [Shamir's secret sharing](https://web.mit.edu/6.857/OldStuff/Fall03/ref/Shamir-HowToShareASecret.pdf) uses polynomial interpolation.
```

threat-model/readme.md
```diff
@@ -230,11 +294,63 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+In terms of our threat model, MPC uses a helper party network composed of aggregators and we assume that an attacker can control some subset of those aggregators. That exact threshold may be different for a given proposal, for example, we may assume that an attacker can only control one out of three aggregators. This would enable, in cryptographic terms, a maliciously secure, honest two out of three majority MPC.
```
csharrison, line 324: Can you please define maliciously secure in the same style you define honest but curious below?

threat-model/readme.md
csharrison, line 322: I think it's fine to mention secret share as an example, but I would prefer that we don't encode that too much into the threat model since there are plenty of MPC protocols we could choose that don't use shares (e.g. shufflers).

threat-model/readme.md
```diff
@@ -230,11 +294,63 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+Given these aggregators, a _client/user_ is able to generate a secret sharing of their input data, and then securely communicate one secret share to each aggregator. This allows the aggregators to then perform the MPC protocol to compute the predefined function. The _client/user_ trusts that an attacker cannot control enough aggregators to violate that protocol.
+
+If the protocol is faithfully executed, the aggregate result can then be reported to the intended recipient without the entities that performed the computation needing to witness the answer.
+
+This threat model does not require that the MPC protocol provide safeguards against an aggregator spoiling the answers from the system. Though a spoiled answer is possible in many protocols, the primary threat that a compromised aggregator presents is to the privacy of _clients/users_. We assume that there is some contractual relationship between those requesting aggregation and the entities that perform that aggregation such that the aggregators lack significant incentive to spoil results.
```
csharrison, line 332:
```suggestion
This threat model does not require that the MPC protocol provide safeguards against an aggregator spoiling the answers from the system. Though a spoiled answer is possible in many protocols, the primary threat that a compromised aggregator presents is to the privacy of _clients/users_. We assume that there is some relationship between those requesting aggregation and the entities that perform that aggregation such that the aggregators lack significant incentive to spoil results.
```
Don't think we need to stipulate on the kind of relationship we assume.

threat-model/readme.md
```diff
@@ -230,11 +294,63 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+### 4.2 Trusted Execution Environments
+
+Trusted execution environments are specialized hardware where encrypted data can be sent "in" to an enclave where it is decrypted and operated on, but cannot be otherwise accessed. A TEE can produce an "attestation" that acts as a claim - backed by the vendor of the TEE - that only specific code is actively running.
```
csharrison, line 336:
```suggestion
Trusted execution environments are specialized hardware where encrypted data can be sent "in" to a computing environment where it is decrypted and operated on, but cannot be otherwise accessed. A TEE can produce an "attestation" that acts as a claim - backed by the vendor of the TEE - that only specific code is actively running.
```
No need to introduce more terms that need defining.

threat-model/readme.md
```diff
@@ -230,11 +294,63 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+Different hardware manufacturers offer different types of TEEs, and there are various assumptions which need to be made about the manufacturer, hardware operator, and other tenants on the hardware in order to achieve our ideal of input data remaining oblivious. There are documented attacks for many of these types of hardware, and while we don't expect all web platform vendors to support this construction, some vendors have expressed comfort with the required assumptions.
```
csharrison, line 338:
```suggestion
Different hardware manufacturers offer different types of TEEs, and there are various assumptions which need to be made about the manufacturer, hardware operator, and other tenants on the hardware in order to achieve our ideal of input data remaining oblivious. There are documented attacks for many of these types of hardware, so different vendors will need to make their own choices about which instantiations (if any) to trust.
```
The threat model doc does not need to speculate about what all vendors will do, imo.

threat-model/readme.md
```diff
@@ -230,11 +294,63 @@ When managing a privacy budget, that budget is assigned to some party who can us
...
+
+In order to utilize a TEE to achieve private computation, we need two properties. First, the data must only be able to be decrypted within the TEE, and second, we need to assure that the TEE only executes code which evaluates the predefined function.
+
+The currently proposed system utilizes a helper party network composed of coordinators. This helper party network participate in a key exchange protocol to generate a _threshold key pair_, with a public key for encryption and several partial private keys such that each coordinator has a single partial private key, and all (or a predefined subset) are required for decryption. This allows the _client/user_ to encrypt the data towards the helper party network.
+
+When a TEE is instantiated to perform an aggregation, each coordinator in the helper party network can validate an attestation from the TEE and provide their partial private key if and only if the attestation verifies that the TEE is only running the expected code. Thus, only the TEE is ever able to decrypt the data, so long as the attacker is only able to control a subset of coordinators in the helper party network.
```
csharrison, line 344:
```suggestion
When a TEE is instantiated to perform an aggregation, each coordinator in the helper party network can validate an attestation from the TEE and provide their partial private key if and only if the attestation verifies that the TEE is running the expected code. Thus, decrypting data outside the TEE is possible only if the attacker is able to control all of the coordinators in the helper party network.
```

threat-model/readme.md
csharrison: nit: s/attribution/measurement, or even just say "capabilities".
Let's keep the threat model as general as possible since we could use it across multiple different proposals.

csharrison (commented):
Thanks @eriktaubeneck for this writeup. In general it looks good to me, although I have one major comment about the min size of coordinator networks.

threat-model/readme.md
```diff
@@ -10,16 +10,24 @@ This document is currently a **draft**, submitted to the PATCG (and eventually t
...
 
-In the presence of this adversary, APIs should aim to achieve the following goals:
+Private computation can be instantiated using several technologies:
+
+* Multi-party computation (MPC) distributes information between multiple independent entities using secret sharing.
+* A trusted execution environment (TEE) isolates computation and its state by using specialized hardware.
+* Fully homomorphic encryption (FHE) enables computation on the ciphertext of encrypted inputs.
 
+Though the implementation details differ for each technology, ultimately they all rely on finding two entities - or _aggregators_ - that can be trusted not to conspire to reveal private inputs. The forms considered by existing attribution proposals are MPC and TEEs.
```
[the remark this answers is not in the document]
eriktaubeneck: agreed

threat-model/readme.md
eriktaubeneck: Yes, I think we need a whole issue / PR to deal with privacy budgeting. It is not at all sufficiently addressed, and, depending on how it works, could present all types of issues.

threat-model/readme.md
```diff
@@ -99,85 +93,159 @@ See [section 2. First Parties, Embedded Parties and Delegated Parties](#2-First-
...
-1. Unencrypted individual share.
-2. All the assets from first and third parties.
-3. Shares of the output of the aggregation protocol.
-4. Identity of other aggregators.
 
+#### 1.5.1. Assets
 
-#### 1.4.2. Capabilities
+1. Unencrypted individual share of a secret share.
+2. All the assets from first and third parties.
```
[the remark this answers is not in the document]
eriktaubeneck: I certainly agree there is some work to do here. I think the idea was that we want to assume that an attacker might not only control some number of helper parties, but also a first or third party (this should also now be delegated party...). From there, we basically assume that the helper party could also have the first/delegated party assets. But, yes, a new issue for this would be great. I would love to get more input here.

I'm unopinionated on party vs site (though maybe we want to use site/app to be explicit.) I'll open a PR with that swap and try to get some input.

threat-model/readme.md
```diff
@@ -99,85 +93,159 @@ See [section 2. First Parties, Embedded Parties and Delegated Parties](#2-First-
...
-We assume the existence of attackers on the network links between various parties.
+1. Coordinators may compromise the availability of the system by refusing to validate the attestation and refusing to supply their key share.
+
+
+#### 1.6.3 Mitigations
+
+1. Helper party networks with a larger number of coordinators could utilize a threshold key which requires less than all key shares to decrypt the data, preventing an individual coordinator from compromising availability. This comes at the cost of more complicated analysis about the ability of an attacker to control different subsets of that helper party network.
+
+
+### 1.7. Helper party collusion
```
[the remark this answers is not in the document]
eriktaubeneck: Probably also between helper parties and first/delegated parties. Will open a new issue.

threat-model/readme.md
```diff
@@ -99,85 +93,159 @@ See [section 2. First Parties, Embedded Parties and Delegated Parties](#2-First-
...
+
+
+### 1.8 Cloud Providers for Helper Parties
+
+Helper parties may run either on physical machines owned by directly by the aggregator or (more commonly) subcontract with a cloud provider. We assume that an attacker can control some subset of cloud providers.
+
+
+#### 1.8.1 Assets
+
+1. All the assets of the helper party(ies) utilizing the cloud provider.
```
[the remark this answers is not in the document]
eriktaubeneck: haha I had to look up what to do here, and there was no good answer (for optional plurals ending in y.) this works for me. also, good tip on utilizing.

threat-model/readme.md
```diff
@@ -99,85 +93,159 @@ See [section 2. First Parties, Embedded Parties and Delegated Parties](#2-First-
...
+
+1. Cloud providers have all the capabilities of the helper parties utilizing that cloud provider.
+2. If a sufficient number of helper parties utilize a common cloud provider,
+    a. for aggregator networks, it can reconstruct the client/user assets;
+    b. for coordinator networks, it can reconstruct the decryption key.
+
+
+#### 1.8.3 Mitigations
+
+1. Helper parties networks should utilize sufficiently distinct cloud providers beyond the proposal specific subset which the attacker is assumed to control.
```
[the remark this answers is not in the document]
eriktaubeneck: My concern here would be if an attacker controls the cloud provider and the first/delegated party, and all coordinators run on that cloud provider, the attacker can construct the entire private key and decrypt the data.

As for getting the key into the TEE, I believe we are assuming a secure communication channel between the coordinator and the TEE (i.e. the TEE has an internal private key with an externally known public key, allowing the coordinator to send in encrypted data that even the TEE operator couldn't see.)

threat-model/readme.md
eriktaubeneck: I'm going to open a new issue, as I'd like to get more input from others. TEEs have multiple different configurations which have different models, and I'm not an expert across them.

threat-model/readme.md
eriktaubeneck: Good callout on the "root" key, will add.

The access to the private key likely differs depending on what type of TEE is used, and I know that the intention in most cases is to not have access to it, but since we can't verify I included it here. Would it help to say something more like "While TEE Manufacturers take measures to prevent the leakage of private keys to themselves, we assume that they may have access to these private keys embedded in the TEEs."?

threat-model/readme.md
eriktaubeneck: Hmm, I'll open a new issue around this, because we also include this up in the threat model part (where we talk about aggregators having access to secret shares.) Maybe we want to be more specific, and then add other types with different analysis as they arise?

threat-model/readme.md
eriktaubeneck, line 324:
```suggestion
In terms of our threat model, MPC uses a helper party network composed of aggregators and we assume that an attacker can control some subset of those aggregators. That exact threshold may be different for a given proposal, for example, we may assume that an attacker can only control one out of three aggregators. This would enable, in cryptographic terms, a _maliciously secure_, honest two out of three majority MPC, where the input data remains oblivious even in the fact of an attacker who controls one of the three aggregators and deviates from the protocol. The protocol would always detect this attacker, and typically aborts the computation.
```

threat-model/readme.md
eriktaubeneck: > all of the coordinators

I was hoping to leave this more flexible, as it seems reasonable to propose a k out of n threshold key, potentially to prevent availability risk.

eriktaubeneck:
```suggestion
When a TEE is instantiated to perform an aggregation, each coordinator in the helper party network can validate an attestation from the TEE and provide their partial private key if and only if the attestation verifies that the TEE is only running the expected code. Thus, decrypting data outside the TEE is possible only if the attacker is able to control a proposal specific subset of coordinators in the helper party network.
```

eriktaubeneck: @csharrison adjusted slightly. LMK if this doesn't make sense to you and we can open a new issue.

csharrison: LGTM but can you hyphenate "proposal-specific"

threat-model/readme.md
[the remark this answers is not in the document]
csharrison: Ah, sorry I missed this assumption. I think what you have makes sense in that case, although I think this is a difference in how the ARA deployment works (which does put root of trust in the cloud operator). I think as long as this is an optional mitigation I am fine with it.

threat-model/readme.md
eriktaubeneck: This relates to what is (newly) included in section 1.4

> All parties in a helper party network should be known a priori and web platform vendors should be able to evaluate risk of an attacker that is more powerful than our assumption, e.g., the attacker is able to control more than the (protocol specific) subset of helper parties in the network.

My contention is that the report collector isn't known a priori, and thus the vendor isn't able to evaluate the risk sufficiently. I'll move this over to a new issue, as this is worth having a more detailed conversation on.

threat-model/readme.md
eriktaubeneck:
```suggestion
Arbitrary helper party networks will not be able to automatically participate in these protocols, as web platform vendors will need to provide some form of authorization. As such, our goal is to reach consensus on a core set of methods and helper party networks so that first/delegated parties can get uniform service across web platform vendors. This will in turn allow sites/apps to access a basic set of standardized capabilities across all participating platforms.
```

threat-model/readme.md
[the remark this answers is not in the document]
eriktaubeneck: It's worth adding a new issue and discussion this assumption directly.

threat-model/readme.md
eriktaubeneck: added to https://github.com/patcg/docs-and-reports/issues/16

threat-model/readme.md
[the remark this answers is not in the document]
eriktaubeneck: Opened #19

threat-model/readme.md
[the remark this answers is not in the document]
eriktaubeneck: Opened #20

threat-model/readme.md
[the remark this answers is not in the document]
eriktaubeneck: Opened #21

threat-model/readme.md
eriktaubeneck: Opened #22

threat-model/readme.md
eriktaubeneck: Opened #23

threat-model/readme.md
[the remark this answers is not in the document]
eriktaubeneck: Opened #24

threat-model/readme.md
eriktaubeneck: Resolving this, but please feel free to open new issues and/or PRs.

threat-model/readme.md
eriktaubeneck: opened #25

eriktaubeneck:
Thank you @martinthomson @bmayd @marianapr @palenica @csharrison for all the feedback. THIS IS STILL A DRAFT, so I'm going to merge it at this point incorporating that feedback. 

We also opened a bunch new issues, where we can continue to discuss some of these individual points:
#15, #16, #17, #18, #19, #20, #21, #22, #23, #24, #25

@AramZS we might have a whole mess of agenda topics for our next meeting 😜

threat-model/readme.md
eriktaubeneck: Opened #26

PR #18, opened 2022-10-21
Add design-dimensions document

csharrison:
I've pulled most of the pieces out of the Google doc:
https://docs.google.com/document/d/1_mq2e_Yuxwh1664S3Ep_yVgOECHqyF27Qiev6UrmQFk/edit#

Aside from the "layering ideas" section which I'm still not sure if it belongs here. It's possible the layering ideas can live alongside each design dimension rather than a separate section.

Tables are fairly unwieldy to edit, so appreciate any suggestions for alternative formatting.


eriktaubeneck:
@csharrison thank you for translating this! Agreed that the tables are difficult to edit, but they do visually present the information quite well!

design-dimensions/README.md
```diff
@@ -0,0 +1,66 @@
...
+|------------------------------------|-----------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Server mediated                    | Whether the API makes use of a server to compute the output                                         | PCM: NoE-ARA: No<br>A-ARA: Yes<br>IPA: Yes<br>SKAN: Yes                                                                                                                                                                                                                                                           |
```
martinthomson, line 25:
```suggestion
| Server mediated                    | Whether the API makes use of a server to compute the output                                         | PCM: No<br>E-ARA: No<br>A-ARA: Yes<br>IPA: Yes<br>SKAN: Yes                                                                                                                                                                                                                                                           |
```

design-dimensions/README.md
```diff
@@ -0,0 +1,66 @@
...
+| Where is budgeting applied         | Where are user contributions bounded within the above scope?                                        | PCM: on-device<br>E-ARA: on-device<br>A-ARA: hybrid on-device and in servers<br>IPA: servers<br>SKAN: on-device                                                                                                                                                                                                   |
+| Scope of privacy budgeting         | If applicable, the axes along which privacy budgeting applies: time epoch, site, campaign, delegate | PCM: Per device/source-destination site pair. Limited in rate by reporting delay and user interaction.<br>A-ARA: Per epoch, source ⇔ destination site pair, device.<br>E-ARA: Per epoch, source ⇔ destination site pair, device<br>IPA: Per week/epoch, per site, per match key.<br>SKAN: Per device / source app |
```
martinthomson, line 29:
```suggestion
| Scope of privacy budgeting         | If applicable, the axes along which privacy budgeting applies: time epoch, site, campaign, delegate | PCM: Per device/source-destination site pair. Limited in rate by reporting delay and user interaction.<br>A-ARA: Per epoch, source ⇔ destination site pair, device.<br>E-ARA: Per epoch, source ⇔ destination site pair, device<br>IPA: Per week/epoch, per site<br>SKAN: Per device / source app |
```

@eriktaubeneck check my work please, but I believe that we've dispensed with a per-match key budget.  We're now looking at query limiting (and maybe public query logs).

martinthomson (approved):

design-dimensions/README.md
csharrison: "match key" here is standing in for the user. How are user contributions bounded in IPA if not by match key?

eriktaubeneck: You are both right. Contributions are bound by match key (as a stand-in for user), but the budget it tracked globally, based on the worst case assumption that every match key is included in every query (which is what @martinthomson is referring to.)

Looking back at this, I'm not sure which the intent is here, but I am assuming the former. It might just be worth calling out that the scope is per match key, but it's tracked by number of queries?

eriktaubeneck:
@csharrison other than the one open suggestion, I'd propose we merge this in (and open a new issue for that one suggestion.)

csharrison:
> @csharrison other than the one open suggestion, I'd propose we merge this in (and open a new issue for that one suggestion.)

Yes I've been planning on it, just needed a prodding :)

csharrison:
Err.. actually I don't think I have write access to this repo so I need someone else to merge.

design-dimensions/README.md
csharrison: The intention with this row was to document abstractly the scope of privacy, in the differential privacy "privacy unit" sense, and less about the technical details about how we do it (usually a combination of contribution bounding and rate limiting).

eriktaubeneck: In that case, I think this is accurate and doesn't need the suggestion. @martinthomson feel free to open a new issue if you think it's worth including the "how".

PR #26, opened 2022-10-21
clarify and expand TEE manufacturer assets

eriktaubeneck:
Small PR to address @martinthomson [comment here](https://github.com/patcg/docs-and-reports/pull/14/files#r1001405464):

> It is quite possible that operational practices at the TEE manufacturer is good enough that they don't have access to the private keys, but they will have access to the "root" key.
>
> We should assume that they have both.

_Originally posted by @martinthomson in https://github.com/patcg/docs-and-reports/pull/14#discussion_r1001405464_

### DIFF
--- a/threat-model/readme.md
+++ b/threat-model/readme.md
@@ -220,7 +220,9 @@ TEEs can provide "attestation" which verifies that the TEE is running in the exp
 
 #### 1.9.1 Assets
 
-1. Private keys embedded in TEEs and their corresponding public keys.
+1. While TEE Manufacturers take measures to prevent the leakage of private keys to themselves, we assume that they may have access to these private keys embedded in the TEEs.
+2. The corresponding public keys for a specific TEE.
+3. The "root" key used to verify the authenticity of a specific public key.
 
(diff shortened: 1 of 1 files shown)


threat-model/readme.md
```diff
@@ -220,7 +220,9 @@ TEEs can provide "attestation" which verifies that the TEE is running in the exp
 
 #### 1.9.1 Assets
 
-1. Private keys embedded in TEEs and their corresponding public keys.
+1. While TEE Manufacturers take measures to prevent the leakage of private keys to themselves, we assume that they may have access to these private keys embedded in the TEEs.
+2. The corresponding public keys for a specific TEE.
+3. The "root" key used to verify the authenticity of a specific public key.
```
martinthomson, line 225: This might instead be:

```suggestion
3. The "root" private key used to authenticate the keys used in each TEE.
```

The "root" public key can be assumed as public knowledge and not worth mentioning here.  It wasn't clear which you meant here, verification being something that only depends on the public key.

martinthomson (commented):
Thanks!

PR #32, opened 2023-02-21
adding tee operator. too much side channel?

kylehogan:
First draft of threat model for the TEE operator. 
Lots of side channel discussion (maybe too much side channel discussion).
Some of these side channels are not currently possible to mitigate.
Do we want references here?
Also see #33 which adds the TEE itself and #34 which adds the TEE manufacturer.

### DIFF
--- a/threat-model/readme.md
+++ b/threat-model/readme.md
@@ -221,8 +221,6 @@ Most commonly, the operator will be a cloud provider.
 4. Physical access mitigations: Defending against power/acoustic/temperature side channels requires preventing the operator from observing these features or by executing TEE processes obliviously (execution is fully independent of the confidential input data).
 5. Denial of service: ensure that the inputs and outputs of the computation remain confidential so that the operator cannot selectively deny service.
 
-
-<a name="tee-manufacturers"></a>
 ### 1.10 TEE Manufacturers
 
 TEEs can provide "attestation" which verifies that the TEE is running in the expected state and running the expected code. These attestations are typically produced with an asymmetric key, where the private key is physically embedded into the TEE, and the public key is published via a system similar to a certificate authority.
--- a/threat-model/readme.md
+++ b/threat-model/readme.md
@@ -200,18 +200,25 @@ Helper parties may run either on physical machines owned by directly by the aggr
 
 ### 1.9 Operators of TEEs
 
-As a piece of hardware, TEEs will have an operator with access to the machine. Most commonly, this will be a cloud provider. Depending on the specific hardware, there may be known vulnerabilities in which an attacker who only controls the operator can violate the obliviousness of client/user data. These attacks are outside this threat model, but are likely to inform specific web platform decisions about which instantiations of private computation to support.
+Trusted Execution Environments (TEEs) offer hardware supported confidential computing.
+The operator of a TEE is the entity who controls the physical host machine on which this hardware is installed.
+Most commonly, the operator will be a cloud provider. 
 
 #### 1.9.1 Assets
-TODO
+1. The host machine with the TEE hardware installed.
 
 
 #### 1.9.2 Capabilities
-TODO
+1. The ability to run arbitrary processes on the host machine. TEEs are designed to provide confidentiality and integrity despite this, but, in practice may be vulnerable to side channel leakage via, e.g., the cache or memory access patterns or speculative/transient execution attacks like Spectre. (see Section 1.10 for more on the capabilities of a TEE.)
+2. Physical access to the host machine. The operator typically has physical access to the machine that hosts the TEE. Thus, it may attempt to compromise confidentiality via a cold boot attack or side channels using power consumption, acoustics, or temperature.
+3. Ability to deny service by restricting access to the TEE. Denial of service can present an issue if the operator is able  *selectively* restrict operation of the TEE based on the output of its computation. 
 
 
 #### 1.9.2 Mitigations
-TODO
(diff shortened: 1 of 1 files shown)


threat-model/readme.md
```diff
@@ -200,18 +200,25 @@ Helper parties may run either on physical machines owned by directly by the aggr
...
 
 #### 1.9.2 Mitigations
-TODO
+1. Microarchitectural side channel mitigations: microarchitectural side channels arise due to resource sharing between the TEE processes and other processes on the host machine. They can be mitigated by ensuring spatial and temporal isolation of the TEE processes though this has proved challenging in practice and no existing TEE is free of side channel leakage.
+2. Speculative/transient execution mitigations: turn off speculative execution. 
+3. Physical access mitigations: To mitigate cold boot attacks any TEE assets stored off the trusted hardware, e.g. in DRAM, must be encrypted and integrity checked. Defending against power/acoustic/temperature side channels requires preventing the operator from observing these features or by executing TEE processes obliviously (execution is fully independent of the confidential input data).
+4. Denial of service: ensure that the inputs and outputs of the computation remain confidential so that the operator cannot selectively deny service.
 
 
 ### 1.10 TEE Manufacturers
```
martinthomson, line 224:
```suggestion
<a name="tee-manufacturers"></a>
### 1.10 TEE Manufacturers
```

threat-model/readme.md
```diff
@@ -200,18 +200,25 @@ Helper parties may run either on physical machines owned by directly by the aggr
...
+2. Physical access to the host machine. The operator typically has physical access to the machine that hosts the TEE. Thus, it may attempt to compromise confidentiality via a cold boot attack or side channels using power consumption, acoustics, or temperature.
+3. Ability to deny service by restricting access to the TEE. Denial of service can present an issue if the operator is able  *selectively* restrict operation of the TEE based on the output of its computation. 
 
 
 #### 1.9.2 Mitigations
-TODO
+1. Microarchitectural side channel mitigations: microarchitectural side channels arise due to resource sharing between the TEE processes and other processes on the host machine. They can be mitigated by ensuring spatial and temporal isolation of the TEE processes though this has proved challenging in practice and no existing TEE is free of side channel leakage.
```
```suggestion
1. Microarchitectural side channel mitigations: microarchitectural side channels arise due to resource sharing between the TEE processes and other processes on the host machine. They can be mitigated by ensuring spatial and temporal isolation of the TEE processes. These mitigations have proven to be challenging in practice and no existing TEE is free of side channel leakage.
```

threat-model/readme.md
```diff
@@ -200,18 +200,25 @@ Helper parties may run either on physical machines owned by directly by the aggr
...
+2. Speculative/transient execution mitigations: turn off speculative execution. 
+3. Physical access mitigations: To mitigate cold boot attacks any TEE assets stored off the trusted hardware, e.g. in DRAM, must be encrypted and integrity checked. Defending against power/acoustic/temperature side channels requires preventing the operator from observing these features or by executing TEE processes obliviously (execution is fully independent of the confidential input data).
```
martinthomson, line 220: This might be split:

```suggestion
3. To mitigate cold boot attacks any TEE assets stored off the trusted hardware, e.g. in DRAM, must be encrypted and integrity checked.
4. Defending against power/acoustic/temperature side channels requires preventing the operator from observing these features or by executing TEE processes obliviously (execution is fully independent of the confidential input data).
```

Though I think that the latter could be:

> Defending against attacks that use power/acoustic/temperature side channels requires either preventing the operator from observing these features by limiting physical access in some way or by executing TEE processes obliviously (that is, eliminating any potential for side channel information by using computation that does not depend on the content of any confidential data).

Out of curiosity, what is the state of the art as it relates to *timing* side channels?  You don't mention them, but these no doubt exist as well and leak information to more than just those with physical access to the machine.

threat-model/readme.md
```diff
@@ -200,18 +200,25 @@ Helper parties may run either on physical machines owned by directly by the aggr
...
+Most commonly, the operator will be a cloud provider. 
 
 #### 1.9.1 Assets
-TODO
+1. The host machine with the TEE hardware installed.
 
 
 #### 1.9.2 Capabilities
-TODO
+1. The ability to run arbitrary processes on the host machine. TEEs are designed to provide confidentiality and integrity despite this, but, in practice may be vulnerable to side channel leakage via, e.g., the cache or memory access patterns or speculative/transient execution attacks like Spectre. (see Section 1.10 for more on the capabilities of a TEE.)
```
martinthomson, line 212: Section numbering is a curse, especially when it is manually maintained like this.  Maybe use `[below](#tee-manufacturers)` and add something for that to point to (see below).

```suggestion
1. The ability to run arbitrary processes on the host machine. TEEs are designed to provide confidentiality and integrity despite this, but, in practice may be vulnerable to side channel leakage via, e.g., the cache or memory access patterns or speculative/transient execution attacks like Spectre. (see [below](#tee-manufacturers) for more on the capabilities of a TEE.)
```

martinthomson (commented):
Thanks for doing this Kyle, This works for me.  Tweaks suggested.

threat-model/readme.md
kylehogan: I think this section in particular is quite rough and needs to be workshopped a bit. 

I separated side channels out into microarchitectural, speculative (a special kind of microarchitectural really), and physical where this is the means an adversary uses to observe something about the program execution. The something observed is very often a timing difference. Maybe there is a better way to describe side channels that makes this more clear? Especially because some of the more traditionally physical attacks can be executed remotely now: [Hertzbleed](https://www.usenix.org/system/files/sec22-wang-yingchen.pdf)

It might also make sense to pull oblivious execution out into its own mitigation as it really applies more generally (but isn't typically a viable option due to the performance).

(agreed on splitting cold boot attacks out)

abo-L (approved):

abo-L (approved):

PR #33, opened 2023-02-21
adding TEE to threat model

kylehogan:
Adding a threat model section for the TEE itself.
Should it have references?
Also see #32 which adds the TEE operator and #34 which adds the TEE manufacturer.


threat-model/readme.md
```diff
@@ -213,6 +213,28 @@ TODO
 #### 1.9.2 Mitigations
 TODO
 
+### 1.10 TEE
+
+TEEs are hardware supported environments for computations that isolate associated processes from other processed on the host machine. TEEs are intended to provide confidentiality, integrity, and authenticity for computations against a, possibly malicious, operator of the host machine.
```
```suggestion
TEEs are hardware supported environments for computations that isolate associated processes from other processed on the host machine. TEEs are intended to provide confidentiality, integrity, and authenticity for computations against a possibly malicious operator of the host machine or - in a shared hosting environment - other users of the same host.
```

threat-model/readme.md
```diff
@@ -213,6 +213,28 @@ TODO
...
+3. Maintain a confidential and integrity checked communication channel between the TEE and the owner of the computation.
+
+#### 1.10.3 Mitigations
+
+1. Confidentiality of attestation/identity key: the TEE's private key(s) should never be accessible from outside the TEE and any operations involving them must be performed obliviously (independently of the key) to avoid leaking bits of the key via side channels.
+3. Integrity of computation: any TEE assets leaving the TEE, e.g. to DRAM, must be integrity checked by the TEE upon return. This includes the initial inputs and final output of the computation over its communication channel with the computation owner.
+4. Confidentiality of computation and its inputs/outputs: guaranteed based on confidentiality of the key(s) associated with the communication channel and resistance to side channel leakage (see 1.9 on TEE operators)
+5. Authenticity of the computation: guaranteed based on confidentiality of the attestation/identity key.
 
 ### 1.10 TEE Manufacturers
```
martinthomson, line 239: For other editors: time to renumber?  Or maybe we just drop numbers entirely.  This is where the limitations of the format stand out.

(Note that the "TEE Manufacturers" section includes subsection 1.**9**.1.

threat-model/readme.md
```diff
@@ -213,6 +213,28 @@ TODO
...
+
+#### 1.10.1 Assets
+
+1. Private attestation/identity key(s) embedded in TEEs by the manufacturer of the TEE. The TEE must maintain confidentiality for the attestation/identity key(s) in order to maintain authenticity of the TEE. That is, if the confidentiality of attestation/identity key(s) is compromised, then a TEE may be impersonated and not provide the expected guarantees.
+2. The computation to be performed. This computation must be correct in that it produces the expected output on given input data and does not leak any input data outside of the TEE. This computation if part of the trusted code/computing base and the TEE cannot provide confidentiality for a computation that directly or indirectly reveals information about its inputs. The TEE is designed to provide integrity for the computation in that it will execute the attested computation as specified. Integrity is based on trust in the manufacturer constructing the TEE hardware according to specification and in integrity checking for any interactions with the host system, e.g. via memory.
```
martinthomson, line 223:
```suggestion
2. The computation to be performed. This computation must be correct in that it produces the expected output on given input data and does not leak any input data outside of the TEE. This computation is part of the trusted code/computing base and the TEE cannot provide confidentiality if a computation directly or indirectly reveals information about its inputs. The TEE is designed to provide integrity for the computation in that it will execute the attested computation as specified. Integrity is based on trust in the manufacturer constructing the TEE hardware according to specification and in integrity checking for any interactions with the host system, e.g., via memory.
```

The last two sentences here seem to be out of place.  I'm wondering if it might make more sense if it were moved to the top section.  Or maybe point 2 in the Capabilities section.  WDYT?

threat-model/readme.md
```diff
@@ -213,6 +213,28 @@ TODO
...
+#### 1.10.2 Capabilities
+
+1. Generate a signature, or "attestation", on the computation using the private attestation/identity key.
+2. Execute the computation as specified with integrity and confidentiality against other processes and the operator.
+3. Maintain a confidential and integrity checked communication channel between the TEE and the owner of the computation.
+
+#### 1.10.3 Mitigations
+
+1. Confidentiality of attestation/identity key: the TEE's private key(s) should never be accessible from outside the TEE and any operations involving them must be performed obliviously (independently of the key) to avoid leaking bits of the key via side channels.
+3. Integrity of computation: any TEE assets leaving the TEE, e.g. to DRAM, must be integrity checked by the TEE upon return. This includes the initial inputs and final output of the computation over its communication channel with the computation owner.
```
martinthomson, line 235:
```suggestion
3. Integrity of computation: any TEE assets leaving the TEE, e.g., to DRAM, must be integrity checked by the TEE upon return. This includes the initial inputs and final output of the computation over its communication channel with the computation owner.
```

threat-model/readme.md
```diff
@@ -213,6 +213,28 @@ TODO
...
+4. Confidentiality of computation and its inputs/outputs: guaranteed based on confidentiality of the key(s) associated with the communication channel and resistance to side channel leakage (see 1.9 on TEE operators)
```
martinthomson, line 236:
```suggestion
4. Confidentiality of computation and its inputs/outputs: guaranteed based on confidentiality of the key(s) associated with the communication channel and resistance to side channel leakage (the section on [TEE operators](#tee-operators) discusses side channels).
```

(For that link to work, you'll need to add an anchor, but I can't make that suggestion here.)

Don't you also need to encrypt any asset that leaves the TEE?  You have integrity checks above, but confidentiality is only assured if those are encrypted also (jointly, that means AEAD as a practical matter).

martinthomson (approved):
Thanks again Kyle.

threat-model/readme.md
kylehogan: good catch on the other users. Do we want to call them users or is there a better term? I also like "colocated" for processes assigned to the same host in cloud environments.

threat-model/readme.md
kylehogan: Happy with moving discussion of integrity to capabilities. Will do.

threat-model/readme.md
kylehogan: Yes! Will make that explicit here.

threat-model/readme.md
eriktaubeneck: fixed

PR #35, opened 2023-02-25
Add new document to address working parameters

eriktaubeneck:
A document to capture the items that we've reached soft consensus on, and to allow for other folks who weren't in attendance to provide input.

cc @AramZS @seanturner


working-parameters/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+At the February 2023 in person meeting in NYC, we ran a [survey and discussed the results](https://github.com/patcg/meetings/issues/91). We were able to discuss two of these topics in detail at the meeting, and intend to continue that conversation in upcoming meetings. As part of those discussions, we reached soft consensus on working parameters that the editors of the _Private Measurement Technical Specification MVP_ can work from.
+
+This document serves to capture those parameters (and is intended to be updated as more are discussed and reach soft consensus.)
+
+
+# Parameters with Soft Consensus
+
+
+## Server Architecture may exists
```
martinthomson, line 13:
```suggestion
## Server Architecture may exist
```

working-parameters/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+The community group has reached soft consensus that the _Private Measurement Technical Specification MVP_ need not be limited to only on-device processing, however there is a very high bar for any server architecture that is included. This included both a high level of security (which the editors will align with the ongoing [Threat Model](https://github.com/patcg/docs-and-reports/tree/main/threat-model) work) as well as ability to explain the privacy and security properties of the system to end users.
+
+
+## Privacy defined by Differential Privacy 
+
+We’ve explored three main definitions of privacy:
+
+
+
```
martinthomson, line 23:
```suggestion
```

martinthomson (commented):
This notion of "soft consensus" is something I'd like to avoid using here.

I think that the general sense of the room was that we'd like to continue working on the assumption that a) a server is OK, and b) that DP is OK.  Neither is a final commitment, but only permission to proceed.

My sense is that we'll need to see something more concrete before we can say we have consensus about something.

eriktaubeneck:
@martinthomson seems I've mixed up "soft consensus" with "rough consensus", but even then, [looking at the definition from the IEFT](https://www.rfc-editor.org/rfc/rfc7282), it doesn't seem appropriate here.

My intent was the capture (and to have a place to continue to document) assumptions that can help the editors proceed, and not imply any final commitment. Let me rephrase this.

working-parameters/README.md
```diff
@@ -0,0 +1,26 @@
+# Working Parameters for Private Measurement Technical Specification MVP
+
+At our September 2022 meeting at TPAC, 8 people signed up to be editors to draft a _Private Measurement Technical Specification MVP._  That group produced [Private Measurement Important Dimensions for Attribution.](https://github.com/patcg/docs-and-reports/tree/main/design-dimensions)
+
+At the February 2023 in person meeting in NYC, we ran a [survey and discussed the results](https://github.com/patcg/meetings/issues/91). We were able to discuss two of these topics in detail at the meeting, and intend to continue that conversation in upcoming meetings. As part of those discussions, we reached soft consensus on working parameters that the editors of the _Private Measurement Technical Specification MVP_ can work from.
```
eriktaubeneck, line 5:
```suggestion
At the February 2023 in person meeting in NYC, we ran a [survey and discussed the results](https://github.com/patcg/meetings/issues/91). We were able to discuss two of these topics in detail at the meeting, and intend to continue that conversation in upcoming meetings. As part of those discussions, we had general agreement on working parameters that the editors of the _Private Measurement Technical Specification MVP_ can use to make progress on the spec. 
```

working-parameters/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+This document serves to capture those parameters (and is intended to be updated as more are discussed and reach soft consensus.)
+
+
+# Parameters with Soft Consensus
```
eriktaubeneck, line 10:
```suggestion
# Parameters with General Agreement
```

working-parameters/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+
+## Server Architecture may exist
+
+The community group has reached soft consensus that the _Private Measurement Technical Specification MVP_ need not be limited to only on-device processing, however there is a very high bar for any server architecture that is included. This included both a high level of security (which the editors will align with the ongoing [Threat Model](https://github.com/patcg/docs-and-reports/tree/main/threat-model) work) as well as ability to explain the privacy and security properties of the system to end users.
```
eriktaubeneck, line 15:
```suggestion
The community group has reached general agreement that the _Private Measurement Technical Specification MVP_ need not be limited to only on-device processing, however there is a very high bar for any server architecture that is included. This included both a high level of security (which the editors will align with the ongoing [Threat Model](https://github.com/patcg/docs-and-reports/tree/main/threat-model) work) as well as ability to explain the privacy and security properties of the system to end users.
```

working-parameters/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Privacy defined by Differential Privacy 
+
+We’ve explored three main definitions of privacy:
+
+1. Information theoretic
+2. K-anonymity
+3. Differential privacy
+
+The community group has reached soft consensus that the _Private Measurement Technical Specification MVP_ should use a definition of privacy based on differential privacy. This does not rule out the use of other definitions of privacy being used in conjunction with differential privacy, however any proposal should aim to achieve a differential privacy guarantee.
```
eriktaubeneck, line 26:
```suggestion
The community group has reached general agreement that the _Private Measurement Technical Specification MVP_ should use a definition of privacy based on differential privacy. This does not rule out the use of other definitions of privacy being used in conjunction with differential privacy, however any proposal should aim to achieve a differential privacy guarantee.
```

martinthomson:
Those suggestions would resolve my concern.

eriktaubeneck:
thanks for the review @martinthomson. suggestions applied.

working-parameters/README.md
```diff
@@ -0,0 +1,26 @@
+# Working Parameters for Private Measurement Technical Specification MVP
```
csharrison, line 1: Can we come up with a term other than "parameters" which invokes things like concrete epsilon levels, etc. Maybe "design dimensions" to align with the other doc?

eriktaubeneck: Sure, I'm fine with "design dimensions". I initially had "assumptions" written down, but that didn't feel quite right either.

bmayd: Is it accurate to refer to them as "constraints"? That's what I consider them and it may be more recognizable, domain-specific term for folks reviewing the doc.

working-parameters/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+This document serves to capture those parameters (and is intended to be updated as more are discussed and reach soft consensus.)
+
+
+# Parameters with Soft Consensus
+
+
+## Server Architecture may exists
+
+The community group has reached soft consensus that the _Private Measurement Technical Specification MVP_ need not be limited to only on-device processing, however there is a very high bar for any server architecture that is included. This included both a high level of security (which the editors will align with the ongoing [Threat Model](https://github.com/patcg/docs-and-reports/tree/main/threat-model) work) as well as ability to explain the privacy and security properties of the system to end users.
```
bmayd: I think rather than "need not be limited to only on-device processing, however there is a very high bar for any server architecture that is included." say something like: "may include server-based dependencies so long as the server architecture meets minimum security and privacy guarantee requirements".

working-parameters/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+We’ve explored three main definitions of privacy:
+
+
+
+1. Information theoretic
+2. K-anonymity
+3. Differential privacy
+
+The community group has reached soft consensus that the _Private Measurement Technical Specification MVP_ should use a definition of privacy based on differential privacy. This does not rule out the use of other definitions of privacy being used in conjunction with differential privacy, however any proposal should aim to achieve a differential privacy guarantee.
```
bmayd: Think the last sentence could be made clearer, suggest:

This does not preclude the use of other privacy definitions in conjunction with differential privacy, however any proposal should support differential privacy guarantees.

working-parameters/README.md
eriktaubeneck: In some cases, the result is a constraint, but it seems like "design dimension" fits a bit better (and those dimensions may or may not be constrained in specific ways.) Does that make sense?

working-parameters/README.md
eriktaubeneck:
```suggestion
The community group has reached soft consensus that the _Private Measurement Technical Specification MVP_ may include server processing dependencies, however any server architechture must achieve both a high level of security (which the editors will align with the ongoing [Threat Model](https://github.com/patcg/docs-and-reports/tree/main/threat-model) work) as well as ability to explain the privacy and security properties of the system to end users.
```

How about this? I would prefer to avoid "minimum requirements", as it could imply that any solution which meets those requirements would be satisfactory. I think to Luke's point in the meeting, these requirements are necessary, but not always sufficient.

working-parameters/README.md
eriktaubeneck:
```suggestion
The community group has reached soft consensus that the _Private Measurement Technical Specification MVP_ should use a definition of privacy based on differential privacy. This does not preclude the use of other privacy definitions in conjunction with differential privacy, however any proposal should aim to provide differential privacy guarantees.
```
Thanks @bmayd, good suggestion. Made one small tweak.

design-dimensions/Dimensions-with-General-Agreement.md
```diff
@@ -0,0 +1,26 @@
...
+
+# Dimensions with General Agreement
+
+
+## Server Architecture may exist
+
+The community group has reached general agreement that the _Private Measurement Technical Specification MVP_ may include server processing dependencies, however any server architecture must achieve both a high level of security (which the editors will align with the ongoing [Threat Model](../threat-model) work) as well as ability to explain the privacy and security properties of the system to end users.
+
+
+## Privacy defined by Differential Privacy
```
csharrison, line 18:
```suggestion
## Privacy defined at least by Differential Privacy
```

This aligns with the caveat in the paragraph

csharrison (approved):
LGTM

PR #37, opened 2023-03-17
Several tentative updates regarding private computation for ML, the usage of aggregator/coordinator terms and minor typos

mvono:
Hello,

In parallel to existing pull requests, please find attached the following complementary updates. The main changes are motivated hereafter.

- In the beginning, only providing aggregation as a potential use-case of private computation (see e.g. the sentence "Private computation can be used to perform aggregation over inputs which, individually, must not be revealed.") might lead to think that the scope of private computation is limited to this use-case. It would be relevant to provide other examples of private computations adtech companies could perform using APIs based on TEEs or MPC. As an example, machine learning (ML) training for campaign optimization use-cases could be performed (albeit not currently proposed by existing APIs). Mentioning the latter use-case is important since it could be proposed in future APIs to offer adtech companies the opportunity to train private ML pipelines on non-aggregated data instead of resorting to ML training techniques taking as inputs noisy aggregated data obtained via Aggregation APIs. 

- Regarding the "Privacy" definition, "receive nothing beyond the intented result" instead of "learn nothing beyond the intented result" seems more appropriate. Indeed, even with a differentially private output, it could still be possible to learn some insights from the output. On the other hand, receiving a DP output allows to be clear on the threat model that is addressed.

- In the beginning, entities (helper parties) on which MPC and TEEs rely are referred to as _aggregators_. In the sequel,  _aggregators_ are only associated with MPC. It might be more appropriate to state from the beginning that helper parties will be either referred to as aggregators for MPC and coordinators for TEE to avoid some misunderstandings.

- As stated in Section 1.8, instantiating helper parties could be done on-premise or via cloud providers. Even if the on-premise solution might be less common a priori, it might be useful to analyze threat models in that case.


threat-model/readme.md
```diff
@@ -10,22 +10,22 @@ This document is currently a **draft**, submitted to the PATCG (and eventually t
...
 
 Private computation can be instantiated using several technologies:
 
 * Multi-party computation (MPC) distributes information between multiple independent entities using secret sharing.
 * A trusted execution environment (TEE) isolates computation and its state by using specialized hardware.
 * Fully homomorphic encryption (FHE) enables computation on the ciphertext of encrypted inputs.
 
-Though the implementation details differ for each technology, ultimately they all rely on finding at least two entities - or _aggregators_ - that can be trusted not to conspire to reveal private inputs. The forms considered by existing attribution proposals are MPC and TEEs.
+Though the implementation details differ for each technology, ultimately they all rely on finding at least two entities (also called helper parties) that can be trusted not to conspire to reveal private inputs. In the sequel, such entities will be referred to as _aggregators_ and _coordinators_ for MPC-based and TEE-based private computations, respectively.
```
csharrison, line 21:
```suggestion
Though the implementation details differ for each technology, ultimately they all rely on finding at least two entities (also called helper parties) that can be trusted not to conspire to reveal private inputs. Such entities will be referred to as _aggregators_ and _coordinators_ for MPC-based and TEE-based private computations, respectively.
```

csharrison: Not sure what is meant by "in the sequel"

threat-model/readme.md
```diff
@@ -10,22 +10,22 @@ This document is currently a **draft**, submitted to the PATCG (and eventually t
...
 
-For our threat model, we assume that an active attacker can control the network and has the ability to corrupt any number of clients, the parties who call the proposed APIs, and some subset of aggregators, when used.
+For our threat model, we assume that an active attacker can control a network of helper parties and has the ability to corrupt any number of clients, the parties who call the proposed APIs, and some subset of aggregators or collectors, when used.
```
csharrison, line 23:
```suggestion
For our threat model, we assume that an active attacker can control the network and has the ability to corrupt any number of clients, the parties who call the proposed APIs, and some subset of aggregators or collectors, when used.
```

csharrison: I think "the network" is correct here as we're talking about an adversary that can arbitrarily read and write network traffic

threat-model/readme.md
```diff
@@ -10,22 +10,22 @@ This document is currently a **draft**, submitted to the PATCG (and eventually t
...
 
 In the presence of this adversary, APIs should aim to achieve the following goals:
 
-1. **Privacy**: Clients (and, more specifically, the vendors who distribute the clients) trust that (within the threat models), the API is purpose constrained. That is, all parties learn nothing beyond the intended result (e.g., a differentially private aggregation function computed over the client inputs.)
-2. **Correctness:** Parties receiving the intended result trust that the protocol is executed correctly. Moreover, the amount that a result can be skewed by malicious input is bounded and known.
+1. **Privacy**: Clients (and, more specifically, the vendors who distribute the clients) trust that (within the threat models), the API is purpose constrained. That is, all parties receive nothing beyond the intended result (e.g., a differentially private aggregation function computed over the client inputs.)
+2. **Correctness:** Parties receiving the intended result have the guarantee that the protocol is executed correctly. Moreover, the amount that a result can be skewed by malicious input is bounded and known.
```
csharrison, line 28: I'm not sure about this one, as this puts a large burden on the protocol itself (e.g. this language might imply that the protocol must be malicious secure in its correctness). I think we should accept protocols which offer weaker notions of correctness than a 100% guarantee.  Maybe we should copy over language from the privacy section here, saying that parties trust (within the threat models) that the protocol is executed correctly?

csharrison (commented):
Overall I think this looks good. I left a couple comments.

threat-model/readme.md
mvono: In the sequel means _In what follows_. It could be replaced by _In the following sections_ if more explicit.

threat-model/readme.md
mvono: Agreed.

threat-model/readme.md
mvono: Agreed. This suggestion will be made in an additional commit. Thank you!

threat-model/readme.md
```diff
@@ -10,22 +10,22 @@ This document is currently a **draft**, submitted to the PATCG (and eventually t
...
+Many of these proposals attempt to leverage the concept of _private computation_ as a component of these purpose-constrained APIs. An ideal private computation system would allow for the evaluation of a pre-defined function (i.e., the constrained purpose,) without revealing any new information to any party beyond the output of that predefined function. For instance, private computation could be leveraged to address two main advertising use-cases namely reporting and campaign optimization. Regarding reporting, private computation could be used to perform aggregation over inputs which, individually, must not be revealed. For campaign optimization, private computation could be used to train a machine learning algorithm over those private inputs without revealing them.
```
martinthomson, line 13: Do we need to condition that last statement a little more?

I say that because most of the ML proposals involve the private computation only forming a part of the training.  That is, you don't execute most of the protocol in private computation, but only those components that operate on the private data.

```suggestion
Many of these proposals attempt to leverage the concept of _private computation_ as a component of these purpose-constrained APIs. An ideal private computation system would allow for the evaluation of a pre-defined function (i.e., the constrained purpose,) without revealing any new information to any party beyond the output of that predefined function. For instance, private computation could be leveraged to address two main advertising use-cases namely reporting and campaign optimization. Regarding reporting, private computation could be used to perform aggregation over inputs which, individually, must not be revealed. For campaign optimization, private computation could be part of a system that trains a machine learning algorithm, without allowing direct access to private information.
```

threat-model/readme.md
```diff
@@ -10,22 +10,22 @@ This document is currently a **draft**, submitted to the PATCG (and eventually t
...
+For our threat model, we assume that an active attacker can control the network and has the ability to corrupt any number of clients, the parties who call the proposed APIs, and some subset of aggregators or collectors, when used.
 
 In the presence of this adversary, APIs should aim to achieve the following goals:
 
-1. **Privacy**: Clients (and, more specifically, the vendors who distribute the clients) trust that (within the threat models), the API is purpose constrained. That is, all parties learn nothing beyond the intended result (e.g., a differentially private aggregation function computed over the client inputs.)
-2. **Correctness:** Parties receiving the intended result trust that the protocol is executed correctly. Moreover, the amount that a result can be skewed by malicious input is bounded and known.
+1. **Privacy**: Clients (and, more specifically, the vendors who distribute the clients) trust that (within the threat models), the API is purpose constrained. That is, all parties receive nothing beyond the intended result (e.g., a differentially private aggregation function computed over the client inputs.)
```
martinthomson, line 27: I think that learning is a better choice than "receive" here.  Receive implies something far more literal.

threat-model/readme.md
```diff
@@ -10,22 +10,22 @@ This document is currently a **draft**, submitted to the PATCG (and eventually t
...
+2. **Correctness:** Parties receiving the intended result trust (within the threat models) that the protocol is executed correctly. Moreover, the amount that a result can be skewed by malicious input is bounded and known.
```
martinthomson, line 28: This clarification is unnecessary.

threat-model/readme.md
```diff
@@ -169,14 +169,15 @@ An coordinator is type of helper party which participates in a helper party netw
 
 ### 1.7. Helper party collusion
 
-If enough helper parties collude (beyond the proposal-specific subset which an attacker is assumed to control), then none of the properties of the system hold. Such scenarios are outside the threat mode.
+If enough helper parties collude (beyond the proposal-specific subset which an attacker is assumed to control), then none of the properties of the system hold. Such scenarios are outside the threat model.
 
 However, we do assume that an attacker can always control at least one helper party. That is, there can be no perfectly trusted helper parties.
 
+### 1.8 On Premise Solutions for Helper Parties
```
martinthomson, line 176: It might be better to open an issue for this rather than provide only the section heading.

threat-model/readme.md
```diff
@@ -10,22 +10,22 @@ This document is currently a **draft**, submitted to the PATCG (and eventually t
...
 
 Private computation can be instantiated using several technologies:
 
 * Multi-party computation (MPC) distributes information between multiple independent entities using secret sharing.
 * A trusted execution environment (TEE) isolates computation and its state by using specialized hardware.
 * Fully homomorphic encryption (FHE) enables computation on the ciphertext of encrypted inputs.
 
-Though the implementation details differ for each technology, ultimately they all rely on finding at least two entities - or _aggregators_ - that can be trusted not to conspire to reveal private inputs. The forms considered by existing attribution proposals are MPC and TEEs.
+Though the implementation details differ for each technology, ultimately they all rely on finding at least two entities (also called helper parties) that can be trusted not to conspire to reveal private inputs. Such entities will be referred to as _aggregators_ and _coordinators_ for MPC-based and TEE-based private computations, respectively.
```
martinthomson, line 21:
```suggestion
Though the implementation details differ for each technology, ultimately they all rely on finding at least two entities that can be trusted not to conspire to reveal private inputs. Such entities will be referred to as _aggregators_ and _coordinators_ for MPC-based and TEE-based private computations, respectively.
```

threat-model/readme.md
```diff
@@ -77,7 +77,7 @@ In this section, we enumerate the potential actors that may participate in a pro
 #### 1.2.3. Mitigations
 
 1. Modification of client assets should be limited by the API interface to only allow for intended modifications.
-2. Use of differential privacy (see [section 3. Aggregation and Anonymization](#3-Aggregation-and-Anonymization)) should be used to prevent
+2. Use of differential privacy (see [section 3. Aggregation and Anonymization](#3-Aggregation-and-Anonymization)) should be used to prevent [This sentence is not finished]
```
martinthomson, line 80:
```suggestion
2. Use of differential privacy (see [Section 3. Aggregation and Anonymization](#3-Aggregation-and-Anonymization)) should be used to protect the contributions of individual users.
```

martinthomson (changes requested):

threat-model/readme.md
mvono: I am in favour of that if "learning" does not refer to any post-processing operation.

threat-model/readme.md
```diff
@@ -200,7 +201,7 @@ Helper parties may run either on physical machines owned by directly by the aggr
 
 ### 1.9 Operators of TEEs
 
-As a piece of hardware, TEEs will have an operator with access to the machine. Most commonly, this will be a cloud provider. Depending on the specific hardware, there may be known vulnerabilities in which an attacker who only controls the operator can violate the obliviousness of client/user data. These attacks are outside this threat model, but are likely to inform specific web platform decisions about which instantiations of private computation to support.
+As a piece of hardware, TEEs will have an operator with access to the machine. Depending on the specific hardware, there may be known vulnerabilities in which an attacker who only controls the operator can violate the obliviousness of client/user data. These attacks are outside this threat model, but are likely to inform specific web platform decisions about which instantiations of private computation to support.
```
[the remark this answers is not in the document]
michaelkleber: I appreciate this change moving away from assuming the TEE operator is a cloud provider, but without at least mentioning them as an example, I think it's hard for a reader to figure out what kind of party is being described.

michaelkleber (changes requested):

threat-model/readme.md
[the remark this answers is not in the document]
AramZS: @mvono @michaelkleber Can you resolve this change before we close this PR?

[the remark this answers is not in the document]
michaelkleber: done

PR #39, opened 2023-03-23
Add a time/interaction budget scope to the general agreement doc

csharrison:
This is coming from a call for consensus in the Mar 23, 2023 meeting.

PR #40, opened 2023-03-23
Add note about rate of information release

martinthomson:
Based on discussion at the 2023-03-23 meeting.  This was what was discussed and (I believe) agreed.  Let's talk about the exact choice of words.


martinthomson:
Closed with #39.

PR #43, opened 2023-05-04
Add private single events to areas of agreement

### LINKED ISSUE #41: Private measurement of single events
I’m posting an issue based on a discussion that happened in https://github.com/patcg-individual-drafts/ipa/issues/60. In that issue, I proposed a new privacy mechanism satisfying differential privacy that works best in the model where individual events are queried, rather than multiple events aggregated together and queried together. While this is a setting where currently both IPA and ARA have no restrictions, we thought it best to bubble up the conversation to this group in general to discuss.

With per-event output, typically results are only useful if you aggregate them in some way after the privacy mechanism / noise is applied (similar in some sense to [local DP](https://en.wikipedia.org/wiki/Local_differential_privacy)), otherwise data is too noisy to do anything useful with it. This is also usually a huge headache because it drowns your data in much more noise than it would normally (e.g. for simple aggregates, noise standard deviation scales with O(sqrt(#users)) vs. O(1)). However, there are a few big reasons why this is useful:
- **Post-processing is “free”**: Adding noise before aggregation means that you can allow for flexible aggregation as a post-processing step, separate from the platform or any trusted server infra. You can always “requery” an already-noised piece of data as many times as you want without paying more privacy budget, because the privacy is “built into” the data. For certain deployments with many downstream use-cases that otherwise would need a lot of requerying, this can end up being a more efficient mechanism.
- **Pushes complexity out**: If the aggregation is very complicated , potentially stateful, or difficult to pull off in MPC or at-scale, then event-level output frees the report collector to be able to implement this themselves. This is particularly useful for things like model training where the “aggregation” step is potentially an entire training pipeline. In fact, the motivation for the mechanism I proposed in https://github.com/patcg-individual-drafts/ipa/issues/60 is to support recent research in private model training, where alternatives like [DP-SGD](https://arxiv.org/abs/1607.00133) are much more complex to support.

The main questions for the group are:
- Are we comfortable with per-event queries that satisfy a similar differential privacy bar as aggregates would? Is differential privacy _alone_ satisfactory for us, or do we need an auxiliary privacy notion to protect against these kinds of queries?
- If not, how should we formalize the kind of protection we want that prevents them? Are these protections realistic within our threat model (sybil attacks, etc)?

Personally, I think we should support this kind of privacy mechanism as long as it satisfies our high level differential privacy bounds, given the benefits listed above, the challenges inherent with protecting this boundary, and the robustness of DP as a definition.

cc @eriktaubeneck @benjaminsavage @bmcase

csharrison:
Fixes #41 

Per the call on May 4 I'll leave this PR open for some time to get feedback.


ShivanKaul:
Given that people using multiple devices/browsers will now not get the same match key, we can't guarantee the contribution limit per-user. Right?

csharrison:
> Given that people using multiple devices/browsers will now not get the same match key, we can't guarantee the contribution limit per-user. Right?

Yeah currently there aren't any measurement proposals that can enforce a sensitivity bound across devices, although it's something that I think we should invest some time into trying to do (e.g. this is possible to some extent with vendor provided match keys). I tried to keep this general so that things like tightening budgets based on estimates of devices per user is a feasible mitigation.

design-dimensions/Dimensions-with-General-Agreement.md
```diff
@@ -25,6 +25,11 @@ We’ve explored three main definitions of privacy:
 
 The community group has reached general agreement that the _Private Measurement Technical Specification MVP_ should use a definition of privacy based on differential privacy. This does not preclude the use of other privacy definitions in conjunction with differential privacy, however any proposal should aim to provide differential privacy guarantees.
 
+### Private measurement of single events
+The group has discussed the use-case of “differentially private measurement of single events” i.e. whether, under a differential privacy guarantee, we would permit queries which ask about a single event’s outcome (did impression lead to a conversion, for instance). For more information about this, see https://github.com/patcg/docs-and-reports/issues/41 and the [agenda topic](https://github.com/patcg/meetings/issues/112) from the May 2023 meeting.
```
martinthomson, line 29:
```suggestion
The group has discussed the use-case of “differentially private measurement of single events”. Specifically, whether, given differential privacy protections, we would permit queries which ask about the outcome for a single event (did impression lead to a conversion, for instance). For more information about this, see https://github.com/patcg/docs-and-reports/issues/41 and the [agenda topic](https://github.com/patcg/meetings/issues/112) from the May 2023 meeting.
```

design-dimensions/Dimensions-with-General-Agreement.md
```diff
@@ -25,6 +25,11 @@ We’ve explored three main definitions of privacy:
...
+
+We have reached a general consensus that these types of queries can be particularly sensitive, so we must take care when setting the proper [differential privacy scope](#privacy-unit--privacy-budget-scoping) as well as privacy parameters (e.g. epsilon) to ensure that events are properly protected, and we ensure that the protection covers people that submit many events. However, queries of this form can be generally permitted, and we do not require special mitigations to ensure aggregation across many events is happening.
```
martinthomson, line 31:
```suggestion
We have reached consensus that these types of queries can be particularly sensitive, so we must take care when setting the proper [differential privacy scope](#privacy-unit--privacy-budget-scoping) and parameters (e.g., epsilon and delta) to ensure that user contributions are adequately protected. Any protection needs to cover people that submit many events. However, queries of this form will be considered.
```

I might be in the rough with this last point, but I'm not giving up on requiring additional protections just yet.  Yes, our analytic framework does not give us the tools necessary to obtain any surety about the results being for more than one person.  That is an excuse, not justification.

martinthomson (changes requested):
Some grammatical nits and maybe a continuation of a prior disagreement.

I want to be constructive here, but I think that memorializing this conclusion in this way is not setting a good precedent for how we treat privacy.

I know that for some people participating here, this is not a huge concern.  They rely on information like this and treat it with the respect that it deserves.  Those people are also rightly concerned that some privacy protections will affect their business.  DP-protected event-level information is super valuable and unspecified additional constraints on the use of that information might make it infeasible to optimize, which is a very important use case.

However, while I regard the general sentiment here favourably, I have to regard the specific wording chosen as too strong of a concession.  Our tools are pretty badly limited here.  DP is great, but this is a topic that shows its limitations.

We already made something of a concession when it came to the continuous release of information.  Maybe that is really the same concession as this one, but I can rationalize the continuous release of information on the basis of that being associated with continuous actions and interactivity with the recipients of the information.  I don't have a similar rationalization to lean on here, so I'm asking that we temper the language a little.

design-dimensions/Dimensions-with-General-Agreement.md
csharrison: > I'm not giving up on requiring additional protections just yet

Is this coming from a concern that DP guarantees will be too weak? I tried to phrase this so that essentially we could agree that plugging in _some_ epsilon could satisfy us. What do you think about something like the following for the last sentence, which carves out mitigations in case we aren't satisfied with the DP.

```suggestion
We have reached consensus that these types of queries can be particularly sensitive, so we must take care when setting the proper [differential privacy scope](#privacy-unit--privacy-budget-scoping) and parameters (e.g., epsilon and delta) to ensure that user contributions are adequately protected. Any protection needs to cover people that submit many events. However, queries of this form will not necessarily be prohibited, although additional protections should be considered if differential privacy guarantees are deemed too weak.
```

I think this ensures that we don't "overly rely" on a DP bound that we don't believe offers enough protection and leaves room for researching mitigations.

martinthomson: That works for me.

martinthomson (approved):
To expand on my earlier point a little (and thanks @csharrison for entertaining a less objectionable alternative; I have not had enough sleep this week and so I'm not sure that I'm coherent)...

People's intuition with aggregation might be considered a rough form of DP.  That is, my contribution is, to some extent, hidden by the noise of others' contributions.  We don't have a strong formalism for that, but that's a useful intuition that we might be able to rely on.  My guess is that this is why aggregated values perform worse: because the noise that the training system experiences is concretely higher when aggregated, even though the formal protections remain the same.

The problem is that many factors (Sybil in particular) conspire to ensure that we will never be able to rely on anything like this formally.  Because that inherent noise is valuable.

If we stick to thinking about DP formally, we are looking at the worst case, which I love, but it means that we only have DP left if we have to contend with all these external factors.  But if we can do something - and it might not be technical in nature - to control for those external factors, then we might be in a stronger position.

csharrison:
> That is, my contribution is, to some extent, hidden by the noise of others' contributions. We don't have a strong formalism for that, but that's a useful intuition that we might be able to rely on. My guess is that this is why aggregated values perform worse: because the noise that the training system experiences is concretely higher when aggregated, even though the formal protections remain the same

This is true for DP-SGD style learning because the noise is applied to the entire gradient in order to keep both the features and labels private, and the gradient can be huge. There could very well be aggregate training techniques in the "label DP" setting which outperform single-event queries, we just don't know of an algorithm yet. Generally speaking aggregation performs _better_ than applying noise to every input because you _don't_ have any noise in others' contributions - you just apply a O(1) noise share to the entire aggregate.

martinthomson:
That sounds like the kernel of a solution to me.  In the general case, maybe you can't ensure that the gradient (or even a given dimension of it) isn't exclusively the result of the contributions of a single user.  But I would be surprised if it wasn't possible to bound the sensitivity to be less than the size of the entire gradient.

design-dimensions/Dimensions-with-General-Agreement.md
```diff
@@ -25,6 +25,11 @@ We’ve explored three main definitions of privacy:
...
 
+### Private measurement of single events
+The group has discussed the use-case of “differentially private measurement of single events”. Specifically, whether, given differential privacy protections, we would permit queries which ask about the outcome for a single event (did impression lead to a conversion, for instance). For more information about this, see https://github.com/patcg/docs-and-reports/issues/41 and the [agenda topic](https://github.com/patcg/meetings/issues/112) from the May 2023 meeting.
+
+We have reached consensus that these types of queries can be particularly sensitive, so we must take care when setting the proper [differential privacy scope](#privacy-unit--privacy-budget-scoping) and parameters (e.g., epsilon and delta) to ensure that user contributions are adequately protected. Any protection needs to cover people that submit many events. However, queries of this form will not necessarily be prohibited, although additional protections should be considered if differential privacy guarantees are deemed too weak.
```
ShivanKaul, line 31:
```suggestion
We have reached consensus that these types of queries can be particularly sensitive, so we must take care when setting the proper [differential privacy scope](#privacy-unit--privacy-budget-scoping) and parameters (e.g., epsilon and delta) to ensure that user contributions are adequately protected. A complication here is that the system might not know that two or more events are connected to a single user while the attacker might. Any protection needs to cover people that submit many events in such a scenario. 
Queries of this form will not necessarily be prohibited, although additional protections will be considered if differential privacy guarantees are deemed too weak.
```

ShivanKaul: Suggested slightly stronger language to call out the threat explicitly, and make it clear that we _will_ work on additional protections if we think that DP is insufficient (and not just call it a day).

csharrison: I feel like this is concern is a topic for the [privacy unit](https://github.com/patcg/docs-and-reports/blob/main/design-dimensions/Dimensions-with-General-Agreement.md#privacy-unit--privacy-budget-scoping) section, which I'd like to have discussions on / consensus separately. Here I only want to say that the multiple events will be protected _within the privacy unit_ we choose. In discussion of the privacy unit, we can reference this issue. I would be happy to file an issue and we could discuss separately how we handle cases like this.

With that framing in mind this is my suggestion for how to clarify the language (and accept your should --> will wording change):
```suggestion
We have reached consensus that these types of queries can be particularly sensitive, so we must take care when setting the proper [differential privacy scope](#privacy-unit--privacy-budget-scoping) and parameters (e.g., epsilon and delta) to ensure that user contributions are adequately protected. Any protection needs to cover people that submit many events within the chosen privacy scope. However, queries of this form will not necessarily be prohibited, although additional protections will be considered if differential privacy guarantees are deemed too weak.
```

What do you think?

martinthomson: I like @ShivanKaul's suggestion in its essence, but I have to agree with @csharrison on the execution.  Unfortunately, we don't have anything about people or users or whatever written down yet.  Presumably that is the next piece to cover.

ShivanKaul: I guess that while single event queries can make the privacy risk worse (and we call this out), the idea that users might not be entirely captured by the privacy-unit is a different topic than what we're discussing here. Your wording seems fine.

csharrison: Thanks @martinthomson and @ShivanKaul . Applied the most recent edit.

AramZS:
It looks like we've come to a settled place on this. I think it is reasonable to merge in at the two weeks point.

AramZS (approved):

PR #44, opened 2023-06-14
Consensus write-up on "Optimization" Use Case

benjaminsavage:
In our last meeting we reached general agreement that the group wants to support the "Optimization" use case. This is a write-up of that agreement.

### DIFF
--- a/measurement-use-cases.md
+++ b/measurement-use-cases.md
@@ -1,0 +1,24 @@
+# Measurement Use Cases
+
+The group is currently working on developing a privacy preserving ad measurement system. 
+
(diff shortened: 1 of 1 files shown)


csharrison (approved):

AramZS:
General agreement among participants in room and on call for F2F.

AramZS:
We are leaving this open as part of an async consensus call. If there are any comments now is the time to air them.

AramZS:
As discussed, these PRs have been open long enough to acquire consensus that they accurately describe our conversations.

PR #50, opened 2023-07-23
Opt-out should be possible and undetectable

### LINKED ISSUE #49: Principle: user opt-out should be undetectable
This avoids problems where sites might retaliate or marginalize users who decide not to participate.

We've talked about this lots already, but I don't know if we wrote it down.

martinthomson:
For framing any discussion we have regarding #49.

Closes #49.


michaelkleber:
We've gotten some pushback about this.  Do we have a position on whether it's sufficient to reveal a "this API is not available" kind of signal as long as there are lots of _other_ reasons for an API's unavailability, _aside from_ the user opting out?

This seems analogous to e.g. probes for device capabilities.  "No you can't have gyroscope sensor access, either because the user said No or because there just is no gyroscope."

AramZS:
@michaelkleber Seeing that you're noting some change on whether we all agree on this I've marked it as needs work and as a candidate for an agenda item in a meeting. If that makes sense to you can you open an issue in meetings?

michaelkleber:
Let's see if we can resolve it in discussion in #49 — maybe agenda space won't be needed.

design-dimensions/Dimensions-with-General-Agreement.md
```diff
@@ -41,3 +41,11 @@ In order to be useful, measurement systems need to produce a continuous release
...
 
 It is worth noting that, at least among existing differential privacy deployments, a time dimension in the privacy unit is very common. See https://desfontain.es/privacy/real-world-differential-privacy.html for a list of examples here.
+
+
+## User Opt-Out
+
+Users must be able to opt out of participation in any measurement activitity.
+
+If a user opts out, then it must be possible for their user agent to hide this fact from sites.
+For example, in an aggregation service, the user agent might provide a zero value to contribute to the aggregate, but this needs to be indistinguishable from a non-zero value.
```
martinthomson, line 51:
```suggestion
For example, in an aggregation service, the user agent might provide a zero value to contribute to the aggregate, but this needs to be indistinguishable from a non-zero value.

In some designs, an opt-out might not be perfectly indistinguishable, but opt out must be indistinguishable from other normal and commonplace conditions.
For example, if opt out in measurement results in returning no information in response to a request, then participating users must also regularly return no information for reasons other than opt out.
```


I don't know if this would address @michaelkleber's concerns, but I will put this here.

michaelkleber: Yup, this definitely resolves my concerns, thank you.

PR #61, opened 2024-07-02
fill out principle on safe but under user control

npdoty:
references to ancillary uses, controlled deidentified data and personal data from privacy principles

### DIFF
--- a/principles/index.html
+++ b/principles/index.html
@@ -29,7 +29,7 @@
   <section id="abstract">
     <p>
       Privacy is essential to the sustainable success of the advertising ecosystem. This document takes up the
-      W3C TAG's <em>Privacy Principles</em> [[?Privacy-Principles]] and specialises them for advertising-related
+      W3C TAG's <em>Privacy Principles</em> [[Privacy-Principles]] and specialises them for advertising-related
       situations.
     </p>
   </section>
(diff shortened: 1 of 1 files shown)


principles/index.html
```diff
@@ -83,6 +83,12 @@ <h3>Measurement</h3>
 
     <section>
       <h4>Measurement should be private for safe, widespread usage, but always be under user control</h4>
+
+      <p>Measurement and attribution are <a href="https://w3ctag.github.io/privacy-principles/#dfn-ancillary-use">ancillary uses</a> with functionality that may rely on broad usage. (Some forms of measurement are done with smaller, opt-in panels that typically provide identifiable data.) In order to be sufficiently private, measurement should not reveal <a href="https://w3ctag.github.io/privacy-principles/#dfn-data">personal data</a>, by, for example, ensuring measurement data is <a href="https://w3ctag.github.io/privacy-principles/#dfn-controlled-de-identified-data">controlled de-identified data</a>.</p>
+
+      <p>Because measurement and attribution involve all kinds of viewing advertisements and a variety of other actions, in a wide range of different contexts, relying on understanding of, expectations about and consent over cross-context recognition as a result of ad measurement would be inappropriate.</p>
```
martinthomson, line 89:
```suggestion
      <p>Measurement and attribution involve all kinds of viewing advertisements and a variety of other actions, in a wide range of different contexts. Relying on understanding of, expectations about, or consent over cross-context recognition as a result of ad measurement would be inappropriate.</p>
```

I really want an Oxford comma here :)

principles/index.html
```diff
@@ -83,6 +83,12 @@ <h3>Measurement</h3>
...
+
+      <p>Users should always be able to control their participation in measurement systems, even with de-identified data. User agents may, for example, provide a setting for the user's general preference about telemetry and aggregate measurement.</p>
```
martinthomson, line 91:
```suggestion
      <p>Users should always be able to control their participation in measurement systems, even when that participation only produces de-identified data. User agents may, for example, provide a setting for the user's general preference about telemetry and aggregate measurement.</p>
```

martinthomson (approved):

npdoty:
Based on discussion at TPAC today, @csharrison or @michaelkleber are concerned that revealing personal data from a particular first-party is expected and acceptable. I'm still not clear on the implications of that, but comments on this PR or alternative language suggestions would be welcome.

michaelkleber:
The subsequent section "Measurement should not significantly enable cross-context recognition" already addresses the re-identification risk of the API.

Maybe the additional protection here is that an aggregate measurement API should not give a caller any _new_ user-specific information?  (Or that any new information needs appropriate DP protection?)

I think the point of the discussion at TPAC was that the API caller can surely already know some information, e.g. the URL of the page where they just called the API, and we want to avoid saying that the API needs to somehow hide that already-known information.

csharrison:
+1 to @michaelkleber . Maybe it would help to understand @npdoty what you were trying to achieve / protect against with this paragraph that isn't already covered, from the perspective of the API?

PR #62, opened 2024-07-15
list Nick Doty as current editor of Private Ad Technologies Principles

npdoty:
and add link to threat model in How This Document Fits In


AramZS:
Def good.

